PR #3, opened 2016-08-21
Fix: Exception not handled by server thread: AWS.NET.SOCKET_ERROR : Send : [32] Broken pipe

stcarrez:
We must protect the call to Send and catch the AWS.Net.Socket_Error
exception that could be raised due to a broken pipe.  If the exception
is not handled, the AWS slot dies with the "Fatal error, slot X is dead now."
message.  By handling the exception we can recover from broken pipe
and close the socket connection.


anisimkov:
I decided to fix it another way
Commit: a08f500f71e6550e3445f88a81780f730c904579

Stephane, Can you please check is it working for you ?

stcarrez:
Ok. I'll update my server and tell you in two or three weeks but I guess it should be OK.

yunushkin88:
Hello I'm sorry but you do not tell me how the you have installed the web server on Linux or Win I have done according to the instructions here http://docs.adacore.com/aws-docs/aws/building_aws.html but to this does not work for a week) and build.cmd not worked,

PR #11, opened 2017-10-03
Support for client-side logging of SOAP calls

sparre-consafe:
This patch allows the user to attach logging call-backs to the generated SOAP client calls.

### DIFF
--- a/src/soap/soap-client-callback.ads
+++ b/src/soap/soap-client-callback.ads
@@ -6,13 +6,11 @@ package SOAP.Client.Callback is
    --  Callback routine types
    type Pre_Call_CB is not null access
       procedure (Connection : AWS.Client.HTTP_Connection;
-                 SOAPAction : String;
                  Request    : SOAP.Message.Payload.Object;
                  Schema     : SOAP.WSDL.Schema.Definition);
 
    type Post_Call_CB is not null access
       procedure (Connection : AWS.Client.HTTP_Connection;
-                 SOAPAction : String;
                  Request    : SOAP.Message.Payload.Object;
                  Response   : SOAP.Message.Response.Object'Class;
                  Schema     : SOAP.WSDL.Schema.Definition);
@@ -20,13 +18,11 @@ package SOAP.Client.Callback is
    --  Default callback routines
    procedure Null_Pre_Call_Callback
       (Connection : AWS.Client.HTTP_Connection;
-       SOAPAction : String;
        Request    : SOAP.Message.Payload.Object;
        Schema     : SOAP.WSDL.Schema.Definition) is null;
 
    procedure Null_Post_Call_Callback
       (Connection : AWS.Client.HTTP_Connection;
-       SOAPAction : String;
        Request    : SOAP.Message.Payload.Object;
        Response   : SOAP.Message.Response.Object'Class;
        Schema     : SOAP.WSDL.Schema.Definition) is null;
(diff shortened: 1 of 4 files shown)


tools/wsdl2aws-generator-stub.adb
```diff
@@ -488,6 +496,14 @@ package body Stub is
          Text_IO.New_Line (Stub_Adb);
       end if;
 
+      if O.Gen_CB then
+         Text_IO.Put_Line
+            (Stub_Adb,
+             "      Post_Call_Callback (Connection, " &
+                """" & To_String (O.Prefix) & Proc & """, " &
+                "Payload, Response, Schema);");
```
TurboGit: note that Schema is a global variable already available to users. I would not pass it here.

src/soap/soap-client-callback.ads
```diff
@@ -0,0 +1,34 @@
+with SOAP.Message.Payload;
+with SOAP.Message.Response;
+
+package SOAP.Client.Callback is
+
+   --  Callback routine types
+   type Pre_Call_CB is not null access
+      procedure (Connection : AWS.Client.HTTP_Connection;
+                 SOAPAction : String;
```
TurboGit: This SOAPAction is part of the Request parameter. Not sure it is needed in the CB, one can get it with

   SOAPAction : constant String := Status.SOAPAction (Request);

tools/wsdl2aws-generator-stub.adb
Knaldgas: Can I just use SOAP.WSDL.Schema.Definition?
The callback might not know which request caused the callback.

TurboGit: No, the Schema variable is generated at the top of the package hierarchy. See root package of a generated code. And, BTW, the Schema variable above passed to Post_Call_Callback is exactly this variable. Note also that the Schema variable contains the definitions for the whole Web Service and is never modified after the initialisation phase (elaboration).

Knaldgas: I see a Schema definition in the top of the package hierarchy of a specific service-set only. We have multiple service-sets aggregated with awsascb, but a common handler for callbacks across these service-sets.

TurboGit: Indeed, I was thinking about this. So let's keep Schema in the callback.

tools/wsdl2aws-generator-stub.adb
```diff
@@ -488,6 +496,14 @@ package body Stub is
          Text_IO.New_Line (Stub_Adb);
       end if;
 
+      if O.Gen_CB then
```
TurboGit: I'm not sure we want to have this feature tied to the -cb option. We can probably want to be able to get traces without having generated the the SOAP callbacks. The traces are in the client part of the code. Maybe a new option -traces would be better. Do you agree?

src/soap/soap-client-callback.ads
sparre-consafe: Probably better to do it like that then.

tools/wsdl2aws-generator-stub.adb
sparre-consafe: I agree, but was too lazy to figure out how to add a new option.  I'll take a look at it.

sparre-consafe:
I'll give it a new try from a dedicated branch.

PR #13, opened 2018-01-05
Support for client-side logging of SOAP calls

sparre-consafe:
This patch allows the user to attach logging call-backs to the generated SOAP client calls.


TurboGit:
Sounds better. But for a topic branch I'd like it to be linear and cleaner.

Also, I would rebase against master. I don't like having merges into topic branches (you have three merges).

The commit log "Changes from Dagrofa/Per" is not really helpful :) And it sounds like a merge, right?

Please let me know if something is not clear.

sparre-consafe:
I can see your point.  I'll make a fresh pull request, with all the changes collected in a single commit.

TurboGit:
I'll propose a rebased and cleaned version of this.

PR #30, opened 2018-06-16
Xosonc

anisimkov:


config/setup/xutil.adb
```diff
@@ -6,7 +6,7 @@
 --                                                                          --
 --                                 B o d y                                  --
 --                                                                          --
---          Copyright (C) 1992-2012, Free Software Foundation, Inc.         --
+--          Copyright (C) 1992-2018, Free Software Foundation, Inc.         --
```
TurboGit: Why? There is no change in this file.

config/setup/xutil.ads
```diff
@@ -6,7 +6,7 @@
 --                                                                          --
 --                                 S p e c                                  --
 --                                                                          --
---          Copyright (C) 1992-2012, Free Software Foundation, Inc.         --
+--          Copyright (C) 1992-2018, Free Software Foundation, Inc.         --
```
TurboGit: Why? There is no change in this file.

anisimkov: I just copied it from gnat sources.

config/setup/xutil.adb
anisimkov: I just copied it from gnat sources

anisimkov: And I think it is a bit better to update then to do not update.

config/setup/xutil.ads
TurboGit: Ok,  but there is no change here so the update is wrong to me. Some projects do update the copyright notice each start of year, AWS does not.

anisimkov:
Ok, I removed xutil.ad? change from commit

anisimkov:
<<Indeed the constant is not currently used in AWS, but it is harmless? Or does it introduce some conflicts?>> No conflicts. Ok abandon this commit if you don't like it. It was just experimenting for me. Not really need.

PR #35, opened 2018-07-15
Synch with original ZLib-Ada repository

anisimkov:
ZLib-Ada original repository ssh://git.code.sf.net/p/zlib-ada/git
was accepted all AWS modifications and returns one back to AWS here.
At this stage include/zlib?* files absolutely equal to zlib-ada
sources after commit 4b89ae14ca71f8584f8202b8f9bf8874f26431a9.

### DIFF
--- a/include/zlib.adb
+++ b/include/zlib.adb
@@ -1,7 +1,7 @@
 ----------------------------------------------------------------
 --  ZLib for Ada thick binding.                               --
 --                                                            --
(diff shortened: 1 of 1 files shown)


TurboGit:
Do you mean that we could use a full Ada implementation for our Zlib support in AWS?

anisimkov:
No, I don't know full Ada deflate/inflate implementation for now. And I think it's too much efforts to create.

PR #37, opened 2018-08-01
Wsdl array record

TurboGit:
Just to be approved. I'll merge on my side (I want the merge commit to be done with --log).

### DIFF
--- a/src/soap/soap-types.adb
+++ b/src/soap/soap-types.adb
@@ -1,7 +1,7 @@
 ------------------------------------------------------------------------------
 --                              Ada Web Server                              --
 --                                                                          --
---                     Copyright (C) 2000-2017, AdaCore                     --
+--                     Copyright (C) 2000-2018, AdaCore                     --
 --                                                                          --
 --  This library is free software;  you can redistribute it and/or modify   --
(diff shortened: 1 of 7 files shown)


anisimkov:
Why Generate arrayType attribute only for RPC binding ?

TurboGit:
Indeed, a thinko it should be done only when using then encoded encoding and not when in literal encoding style.

TurboGit:
New code in review. Thanks.

src/soap/soap-types.adb
```diff
@@ -1585,11 +1585,14 @@ package body SOAP.Types is
...
+
+      if Encoding = WSDL.Schema.Encoded then
+         Append (Result, " " & SOAP_Enc & ":arrayType=""");
+         Append (Result, Array_Type);
+         Append (Result, '[');
+         Append (Result, AWS.Utils.Image (Natural (O.O'Length)));
+         Append (Result, "]""");
+      end if;
 
       if Encoding = WSDL.Schema.Encoded then
```
anisimkov: 2 same if condition one by one

TurboGit:
Indeed, good catch!

TurboGit:
New code in review.

PR #41, opened 2018-09-13
Add a second time-zone format for a date/time.

TurboGit:
For R913-025.


TurboGit:
In fact your idea is a good one. New PR proposed, easier to read.

PR #43, opened 2018-09-17
awsascb produces invalid Ada code when called without parameters.

Rombobeorn:
When no command line parameters are given, awsascb does not terminate as intended, but proceeds to write Ada files where the parameter-dependent parts are missing.

It also writes the error message to the standard output stream, instead of the standard error stream where error messages should be written.


TurboGit:
The license agreement was never signed. Closing, please feel free to reopen a new PR if needed. And sorry for the late answer!

PR #44, opened 2018-09-22
Changed gpr files to call gnat12

plestedr:
Libre version only runs gnat12.
Some unrelated compilation warnings still exist, will be addressed in another pull.


TurboGit:
Thanks!

PR #58, opened 2018-10-12
Po/server identity

TurboGit:


TurboGit:
But we have User_Agent in all call taking a connection, so for consistency I'm added one in the other routine. Note that the internal object does have a User_Agent field. And we do use that. So the only way without a new parameter would have been to check the user-agent in the header list, if present to use the value to initialize the object and to duplicate the header with user-agent removed. All this is too complex, and again since we have User_Agent for all routine with an HTTP_Connection.

PR #60, opened 2018-10-24
Fixes for websocket clients

briot:
Two minor fixes for websocket clients, discovered while testing various servers


TurboGit:
Thanks Manu!

PR #63, opened 2018-10-31
Fix for RA22-003

anisimkov:


regtests/0317_hide_id/hide_id.adb
```diff
@@ -191,7 +168,9 @@ begin
    Last := Hello'Last - Utils.Image (Server.Status.Port (WS))'Length;
 
    Put_Line (DS, "=============== First");
-   Result := Client.Get (Server.Status.Local_URL (WS) & Hello (1 .. Last));
+   Result := Client.Get
```
TurboGit: This defeat the test, the goal here is to ensure that when nothing is passed we properly get the User_Agent. This is the default behavior and should be checked.

anisimkov: The length adjustment was on all packets but have to be only when default user agent there.

PR #75, opened 2018-12-10
Use GNAT IPv6 functionality.

anisimkov:
R710-016 MA31-001

Add IPv6_Only server flag.


anisimkov:
Maybe we need to support compartibility with older GNAT versions

PR #80, opened 2018-12-23
Fix websocket lock

TurboGit:
Please do not merge from GitHub. Thanks.

### DIFF
--- a/src/core/aws-net-websocket-registry.adb
+++ b/src/core/aws-net-websocket-registry.adb
@@ -145,7 +145,9 @@ package body AWS.Net.WebSocket.Registry is
         (To          : Recipient;
          Message     : Unbounded_String;
          Except_Peer : String;
-         Timeout     : Duration := Forever);
+         Timeout     : Duration := Forever;
+         Error       : access procedure (Socket : Object'Class;
+                                         Action : out Action_Kind) := null);
       --  Send the given message to all matching WebSockets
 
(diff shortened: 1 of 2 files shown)


TurboGit:
What do you mean by should not timeout at all?

anisimkov:
I mean we should now wait any time inside of protected call for the output buffer availability. The right way is non-blocking send outside of protected operations. In this send we can detect that output buffer is full, in that case not sent data should be kept somewhere next to socket and the socket with the data should be placed to some special task waiting on set of such sockets for the output buffer availability. Way back to normal send is another story.

TurboGit:
But this is not what is needed here. This is a blocking Send routine. This is important to ensure the same message is delivered to all waiting sockets. I'm not against a non blocking one, but this is not the semantic of this Send routine.

anisimkov:
I propose to discuss it in slack chat, if you don't mind.

anisimkov:
I removed check for Pre => Host'Length > 0; because empty host has meaning for server, it is bind to all available local addresses. Sometime we can bind to specific IPv4 address but additional bind going to be for all IPv6 addresses. I used this method on some regtest as I can remember.

TurboGit:
> The check "Set.Length > 0" should be before wait and do not need to wait on empty set.

Indeed, we can even removed this condition. We cannot be there if Set.Length = 0.

PR #81, opened 2019-01-16
Fix comment about NETLIB setup option

anisimkov:
MA31-001


anisimkov:
Pascal, I think we should rebase before merge. I don't think we need "merge" commits.

TurboGit:
Sure, for a single commit we can rebase.

anisimkov:
If commits in same pull requrest but do not have functional relation it can be rebased too I think.

TurboGit:
I agree too in this case. But most probably it should have been 2 PR instead of 1 from the beginning.

PR #85, opened 2019-01-22
Bind to IPv6 if possible in NETLIB=ipv6 mode

anisimkov:
Fixes regression 0319_soap_log in NETLIB=ipv6 mode

### DIFF
--- a/config/src/aws-net-std__ipv6.adb
+++ b/config/src/aws-net-std__ipv6.adb
@@ -84,9 +84,18 @@ package body AWS.Net.Std is
      (Host   : String;
       Port   : Natural;
       Family : Family_Type;
-      Flags  : C.int := 0) return OS_Lib.Addr_Info_Access;
+      Flags  : C.int := 0) return not null OS_Lib.Addr_Info_Access;
    --  Returns the inet address information for the given host and port.
    --  Flags should be used from getaddrinfo C routine.
+
+   function Select_IPv6_If_Present
+     (Chain : not null OS_Lib.Addr_Info_Access)
+      return not null OS_Lib.Addr_Info_Access;
+   --  Choose IPv6 address if it exists in the chain. This function is needed
+   --  because getaddrinfo returns IPv4 addresses first for Passive flag in
+   --  hints and IPv6 addresses first for calls without Passive flag. Without
+   --  this call the server is going to be bound to IPv4 address even if IPv6
+   --  local address exists.
 
    function Get_Int_Sock_Opt
      (Socket : Socket_Type; Name : C.int) return Integer;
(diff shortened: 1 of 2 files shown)


config/src/aws-net-std__ipv6.adb
```diff
@@ -88,6 +88,14 @@ package body AWS.Net.Std is
    --  Returns the inet address information for the given host and port.
    --  Flags should be used from getaddrinfo C routine.
 
+   function Choose_IPv6
+     (Chain : OS_Lib.Addr_Info_Access) return OS_Lib.Addr_Info_Access;
```
TurboGit: Looking at the code, it seems that the Chain and return value should never be null, right? If so:

  (Chain : not null OS_Lib.Addr_Info_Access) return not null OS_Lib.Addr_Info_Access;

config/src/aws-net-std__ipv6.adb
```diff
@@ -88,6 +88,14 @@ package body AWS.Net.Std is
...
+   --  Choose IPv6 address if exists in the chain. This function needed because
```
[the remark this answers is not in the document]
anisimkov: done

config/src/aws-net-std__ipv6.adb
```diff
@@ -88,6 +88,14 @@ package body AWS.Net.Std is
    --  Returns the inet address information for the given host and port.
    --  Flags should be used from getaddrinfo C routine.
 
+   function Choose_IPv6
```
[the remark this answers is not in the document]
anisimkov: done

config/src/aws-net-std__ipv6.adb
anisimkov: done

config/src/aws-net-std__ipv6.adb
```diff
@@ -190,8 +199,9 @@ package body AWS.Net.Std is
    is
       use type C.int;
 
-      Info  : constant OS_Lib.Addr_Info_Access :=
+      Raw   : constant OS_Lib.Addr_Info_Access :=
```
TurboGit: So, both constants below can be "not null OS_Lib.Addr_Info_Access :="

anisimkov: I think it is superfluous because "not null" already checked as function result and could not be changed because of "constant". But .. if you prefer, Done.

anisimkov: I told not about that compiler could do not do the superfluous check. I told that we do not need to write "not null" for the constants taking the data from such functions.

anisimkov:
Optimized IPv6 prioritization in NETLIB=gnat mode in separate commit. I think both commits can be rebased before apply.

PR #88, opened 2019-01-25
(Send): avoid extra conversion to unbounded_string

briot:
This requires extra malloc and memcpy, which slows things down


TurboGit:
Sounds good even if i always try to avoid GNAT specific extension. Thanks for the contribution.

PR #93, opened 2019-02-04
Make deflate priority last in Accept-Encoding default header

anisimkov:
S201-013

Some web servers provide deflate compressed body wrong.
See: https://stackoverflow.com/questions/3932117
Detected in https://api.cognitive.microsofttranslator.com/translate

### DIFF
--- a/regtests/0317_hide_id/test.out
+++ b/regtests/0317_hide_id/test.out
@@ -6,7 +6,7 @@ Connection: Close
 X-Request: 1st-First
 Host:
 Accept: text/html, */*
-Accept-Encoding: deflate, gzip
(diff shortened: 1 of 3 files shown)


regtests/0317_hide_id/test.out
```diff
@@ -1,77 +1,77 @@
-=============== First
```
TurboGit: Why is this file fully rewritten? Change of line ending?

anisimkov: Yes, you can see dark green and dark red background where are the real modifications.

anisimkov:
> Just a question. 
Answer is at appropriate conversation.

PR #99, opened 2019-02-28
po/asynch send

TurboGit:


anisimkov:
Looks like the commit is not completed. Where is the new config parameter procesing ? Where is new Asynch Send parameter procesing ? Only in Pre condition ?

anisimkov:
Looks like not all modifications shown in Web interface. I saw complete diff in the https://github.com/AdaCore/aws/pull/99.diff

PR #100, opened 2019-03-02
Fix 0260 regression test directory name

anisimkov:


TurboGit:
Is that just for consistency or does it fixes something? Just curious...

anisimkov:
Just consistency. Fixes nothing.

PR #114, opened 2019-07-24
Use GNAT.Calendar.Time_IO.Value to parse time

anisimkov:
S723-002

### DIFF
--- a/src/soap/soap-types.adb
+++ b/src/soap/soap-types.adb
@@ -29,7 +29,6 @@
 
 pragma Ada_2012;
 
-with Ada.Calendar.Time_Zones;
 with Ada.Float_Text_IO;
 with Ada.Long_Float_Text_IO;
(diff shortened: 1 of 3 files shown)


src/soap/soap-types.adb
```diff
@@ -790,58 +789,8 @@ package body SOAP.Types is
...
-            return '+' & I2D (H) & ':' & I2D (M);
-
-         else
-            return '-' & I2D (H) & ':' & I2D (M);
-         end if;
-      end Image;
-
    begin
-      return GNAT.Calendar.Time_IO.Image (O.T, "%Y-%m-%dT%H:%M:%S")
-        & Image (Calendar.Time_Zones.UTC_Time_Offset (O.T));
```
TurboGit: Why have you removed this part? I mean we do not have proper time offset now. "Z" is only one option.

PR #117, opened 2019-08-22
po/rework nil values

TurboGit:


TurboGit:
> How about True, or trUE, is this values unlegal ?

The RFC always talk about true IIRC and this has always worked since AWS inception. So even on interoperability side it seems ok.

I'll fix the commit log.

TurboGit:
@anisimkov  : Ready for your review.

TurboGit:
Thanks, now manually merged.

PR #121, opened 2019-09-08
0245_errcon/test.opt: Move linux-rhES7 match before 'gnat'

brobecke:
'gnat' is a discriminant which is present when running in the linux-rhES7
configuration, so the corresponding line was selected, because it was
ahead of the line added to match 'linux-rhES7'. Moving it ahead avoids
that issue.

For S806-014


TurboGit:
I see, thanks.

PR #122, opened 2019-09-10
0245_errcon/test.opt: test-rhES7.out only applies to rhES7-x86

burratoo:
Red Hat Linux 7 produces different output on x86 compared to other Linux
variants (including Red Hat Linux 7 x86_64).

For S806-014


TurboGit:
This should have been rebased as we have a single commit. And when I merge (when there is multiple commits) I do merge manually with the log.

burratoo:
> This should have been rebased as we have a single commit. And when I merge (when there is multiple commits) I do merge manually with the log.
> 
My bad, still getting use to using GitHub.

PR #124, opened 2019-09-24
Synchronize with zlib-ada repository

anisimkov:


include/zlib.ads
```diff
@@ -129,7 +129,7 @@ package ZLib is
 
    type Filter_Type is tagged limited private;
    --  The filter is for compression and for decompression.
-   --  The usage of the type is depend of its initialization.
+   --  The usage of the type is depend on its initialization.
```
TurboGit: typo:
--  The usage of the type depends on its initialization

anisimkov: Done

PR #130, opened 2019-11-26
po/soap duration

TurboGit:


TurboGit:
> I only found not capitalized second part of identifier in test Long_float. Not sure it is a problem for test.

I have fixed the case of float, looks better. You need to review again. Thanks.

PR #133, opened 2020-01-17
Add support for sub-seconds in xsd:duration.

TurboGit:
This adds support for sub-seconds in xsd:duration runtime support. The
encoding and decoding of an xsd:duration has been enhanced.

Add corresponding regression test.

For SB14-013.


TurboGit:
Change made, thanks !

TurboGit:
Thanks, now merged.

PR #135, opened 2020-01-30
Fix support for null xsd:duration.

TurboGit:
At least one number must be specified when the duration is 0. In this
case we do output P0Y as a special case.

Continued work for SB14-013.


TurboGit:
Thanks !

PR #136, opened 2020-02-09
Pass correct From_Client to Send procedure

reznikmm:
It seems From_Client is lost for binary web socket messages


TurboGit:
As @anisimkov I prefer #137 which actually remove codes. Will close this one. Thanks.

PR #142, opened 2020-03-25
Fix SOAP serialization to not loose precision.

TurboGit:
Update regression test to cover those cases.

Continued work for T310-007.


TurboGit:
Right, will fix.

TurboGit:
Thanks !

PR #146, opened 2020-04-16
Rework the session startup to avoid a potentially blocking operation.

TurboGit:
Indeed, starting a task from a protected object is a potentially
blocking operation and is detected when using the global
configuration pragma Detect_Blocking.

Fixes T415-011.

### DIFF
--- a/src/core/aws-session-control.adb
+++ b/src/core/aws-session-control.adb
@@ -1,7 +1,7 @@
 ------------------------------------------------------------------------------
 --                              Ada Web Server                              --
 --                                                                          --
---                     Copyright (C) 2000-2012, AdaCore                     --
+--                     Copyright (C) 2000-2020, AdaCore                     --
 --                                                                          --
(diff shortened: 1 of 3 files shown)


briot:
I like the simple patch here, better than the one I had suggested. Thanks

TurboGit:
> I like the simple patch here, better than the one I had suggested. Thanks

Thanks for the review. I'll let you know when it is validated and integrated.

TurboGit:
> No need to have Check_Interval and Lifetime parameters in protected Start now. Can assign it outside.

We have a Get_Lifetime public routine that could be called at anytime and so if not initialized into the PO we have a small window where the Lifetime passed to Start won't be set. So I think we should keep the code as-is.

TurboGit:
> The reading is not protected anyway. So it does not matter either write is protected or not. We can easy reach single write together with starting task. But anyway, the code is not becoming less safe.

But we cannot go the path where both settings Check_Interval and Lifetime are not set in an atomic way. Thanks.

PR #147, opened 2020-04-16
Pass ADAFLAGS to include.gpr and add CPPFLAGS to CFLAGS.

asarhaddon:
Pass ADAFLAGS to include.gpr and add CPPFLAGS to CFLAGS.

This patch does not apply the common style flags to include/*, it only
allow explicit user flags (ADAFLAGS) to override default flags.

The External is in shared.gpr so that there is only one environment
lookup for each variable.

### DIFF
--- a/include/include.gpr
+++ b/include/include.gpr
@@ -64,6 +64,10 @@ library project Include is
          when others =>
             null;
       end case;
+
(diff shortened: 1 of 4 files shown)


TurboGit:
Thanks! Looks good to me. Merging.

PR #153, opened 2020-06-04
po/disable prog ini

TurboGit:


TurboGit:
> I think it would be better to be able to disable reading the aws.ini too.

That's not possible. See the T530-009 discussion. Fact is that the .ini are read during elaboration and there is no way to disable aws.ini before it is read.

PR #156, opened 2020-06-20
Update expired certificates for regression test

anisimkov:
GNULTS 0231_cert_status test output became better.


TurboGit:
Looks ok, can you give me the command used to extend the certificate date, I'd like to add this information into demos/cert/README. Or if you prefer just do it as part of this PR.

regtests/0231_cert_status/create-certificates
```diff
@@ -58,7 +58,7 @@ openssl ca -in aws-server.csr -days $DAYS -cert CA-srv.crt \
```
TurboGit, line 58: The comment at the top of this file seems truncated. Also "for mo than" I suppose should be "for more than" ?

anisimkov: depend on certificate. Certificates for this test extended by the script

anisimkov: The number of days the certificates going to be valid defined on siging this way 
openssl ca -in aws-client.csr -days $DAYS -cert CA-clt.crt \
  -keyfile CA-clt.key -batch -out aws-client.crt

TurboGit: Sure, but you did not regenerate the certificate I suppose, you just extended it, no?

anisimkov: I use to regenerate it completely. Maybe we can keep the same certificate request file. But I did not try it. I think no need to keep it.

anisimkov: As I understanr the procedure. User creates certificate request file and related pair of keys. Sending certificate request with public key to authority. Authority generates signed certificates and sending it back to certificate user. User installing the certificate to own server.

anisimkov: The comment in file was completely wrong, looks like it was part of script where it is copied from. I'll think what to do with it.

anisimkov: Comment fixed.

PR #157, opened 2020-06-23
Fix C compilation warning

anisimkov:
wrappers.c:97:28: warning: passing argument 1 of 'SSL_set_tmp_rsa_callback'
  from incompatible pointer type [-Wincompatible-pointer-types]

This mistake was inherited from OpenSSL HTML documentation page
https://www.openssl.org/docs/man1.0.2/man3/SSL_set_tmp_rsa_callback.html
The documentation in OpenSSL sources is right.

### DIFF
--- a/regtests/0239_priority/priorities.adb
+++ b/regtests/0239_priority/priorities.adb
@@ -68,12 +68,10 @@ procedure Priorities is
    Previous_Chipher : ASU.Unbounded_String;
 
    Ciphers : ASU.Unbounded_String :=
-               To_Unbounded_String
-                 (if GNUTLS
-                  then "NORMAL"
-                       & (if Net.SSL.Version > "GNUTLS 3.6.3"
-                          then ":-VERS-TLS1.3" else "")
-                  else "DEFAULT");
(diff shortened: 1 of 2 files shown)


regtests/0239_priority/priorities.adb
```diff
@@ -68,12 +68,10 @@ procedure Priorities is
...
    Ciphers : ASU.Unbounded_String :=
-               To_Unbounded_String
-                 (if GNUTLS
-                  then "NORMAL"
-                       & (if Net.SSL.Version > "GNUTLS 3.6.3"
-                          then ":-VERS-TLS1.3" else "")
-                  else "DEFAULT");
+               To_Unbounded_String (if GNUTLS then "NORMAL" else "DEFAULT");
+
+   No_TLS13 : constant ASU.Unbounded_String
```
TurboGit, line 73: Just a formatting point, see Ciphers above, the := should be end of line and the To_Unbounded_String indented unto constant (n of constant).

anisimkov: ;= location fixed, It is for No_TLS13 actually. The Ciphers can't be constant because we modify it below.

Issue #138, opened 2020-02-25
WebSocket implementation is buggy or bad documented.


reznikmm:
Hello,
I have a problem with Web_Socket API in AWS (server side). The `AWS.Net.WebSocket` has this:
```
   function Create
     (Socket  : Socket_Access;
      Request : AWS.Status.Data) return Object'Class
   with Pre => Socket /= null;
   --  Create a new instance of the WebSocket, this is used by AWS internal
   --  server to create a default WebSocket if no other constructor are
   --  provided. It is also needed when deriving from WebSocket.
   --
   --  This function must be registered via AWS.Net.WebSocket.Registry.Register

```

Looks like to create a custom web socket object one needs to define a function with the same profile and register it with the `Register` procedure. One (like me) could think that this function will be called to construct a custom web socket when a http client call an URL. AWS indeed calls this, but then it will destroy the object. So if you keep the object to send messages to the client it will raise `PROTOCOL_ERRORaws-net-buffered.adb:147 access check failed` with traceback like:

```
#5  0x00007fc950be087f in <__gnat_rcheck_CE_Access_Check> (file=..., line=...) at a-except.adb:1208
#6  0x00007fc951ce0f82 in aws.net.buffered.get_write_cache (socket=..., <get_write_cacheL>=...) at /tmp/openwrt/aws/src/core/aws-net-buffered.adb:147
#7  0x00007fc951ce2dff in aws.net.buffered.write (socket=..., item=...) at /tmp/openwrt/aws/src/core/aws-net-buffered.adb:440
#8  0x00007fc951d3cf34 in aws.net.websocket.protocol.rfc6455.send_frame_header (protocol=..., socket=..., opcd=..., data_length=..., has_mask=..., mask=...) at /tmp/openwrt/aws/src/core/aws-net-websocket-protocol-rfc6455.adb:767
#9  0x00007fc951d3c903 in aws.net.websocket.protocol.rfc6455.send (protocol=..., socket=..., data=..., from_client=...) at /tmp/openwrt/aws/src/core/aws-net-websocket-protocol-rfc6455.adb:626
#10 0x00007fc951da2a62 in aws.net.websocket.send (socket=..., message=..., is_binary=...) at /tmp/openwrt/aws/src/core/aws-net-websocket.adb:597
```
Property named `C` is reset by `Finalize`:

```
(gdb) bt
#0  aws.net.finalize (socket=...) at /home/max/net/aws/src/core/aws-net.adb:137
[...]

PR #159, opened 2020-06-25
Wip/websocket setup

briot:
A fix for #138, which I have been using for 2 years now. It preserves the websocket object created by the user.

### DIFF
--- a/src/core/aws-server-http_utils.adb
+++ b/src/core/aws-server-http_utils.adb
@@ -37,6 +37,7 @@ with Ada.Strings.Fixed;
 with Ada.Strings.Maps;
 with Ada.Strings.Unbounded;
 with Ada.Text_IO;
+with Ada.Unchecked_Deallocation;
 
 with GNAT.MD5;
 with GNAT.OS_Lib;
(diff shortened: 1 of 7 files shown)


briot:
Hello,
any news on this patch ? it seems to fix the issue reported by Maxim in #138, but I haven't seen any conversation here ?

TurboGit:
@briot : Sorry I read the WIP in the header as a non finalized PR and I expected that you would amend it somehow. Will have a look soon.

src/core/aws-server-http_utils.adb
```diff
@@ -1716,7 +1732,11 @@ package body AWS.Server.HTTP_Utils is
                         Send_WebSocket_Handshake_Error
                           (Messages.S403,
                            Exception_Message (E));
+
+                        --  ??? Is the socket automatically unregistered
```
TurboGit, line 1736: No it is not automatically unregistered, you need to do it.

briot: I changed the code to unregister when needed. The behavior of `Registry.Close` is a bit surprising because it receives an `Object` as parameter, but will internally lookup the pointer and free it. So the pointer I am passing (`WS`) is freed underneath although I never passed it. But I think this is related to #209 so let's not worry about that here

briot:
I changed the patch to properly unregister the websocket as needed. The current behavior of `Registry.Close` is a bit surprising because it receives an `Object` and will internally lookup the corresponding pointer to free it (so although I pass `WS.all`, it actually frees `WS`). But I think this will be changed in the context of #209 anyway, so not important here.

PR #160, opened 2020-06-25
(websockets): remove unneeded conversions to and from unbounded_string

briot:
This improves performance, by avoiding some memory allocations and
 copying of messages.

### DIFF
--- a/src/core/aws-net-websocket-registry.adb
+++ b/src/core/aws-net-websocket-registry.adb
@@ -47,7 +47,6 @@ with AWS.Net.Memory;
 with AWS.Net.Poll_Events;
 with AWS.Net.Std;
 with AWS.Net.WebSocket;
-with AWS.Translator;
 with AWS.Utils;
 
 package body AWS.Net.WebSocket.Registry is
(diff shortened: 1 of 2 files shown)


src/core/aws-net-websocket-registry.adb
```diff
@@ -1409,19 +1364,28 @@ package body AWS.Net.WebSocket.Registry is
...
 
    procedure Send
      (Socket       : in out Object'Class;
       Message      : Stream_Element_Array;
       Is_Binary    : Boolean := True;
       Timeout      : Duration := Forever;
-      Asynchronous : Boolean := False) is
+      Asynchronous : Boolean := False)
+   is
+      A : String (Natural (Message'First) .. Natural (Message'Last))
```
TurboGit, line 1384: This may not been working in Character is not the size of a byte. This is way in Translator.Conversion we have a fast version (doing what you've done there) or a slow version doing the conversion.

src/core/aws-net-websocket.adb
```diff
@@ -586,43 +590,35 @@ package body AWS.Net.WebSocket is
    procedure Send
      (Socket    : in out Object;
       Message   : String;
-      Is_Binary : Boolean := False) is
+      Is_Binary : Boolean := False)
+   is
+      A : Stream_Element_Array
```
TurboGit, line 595: likewise

src/core/aws-net-websocket-registry.adb
briot: At this stage, I don't think there is a platform supported by GNAT where a character is not a byte. So we could add a pragma Assert just to make sure.
The above is more efficient than AWS.Translator, because To_String returns the result on the secondary stack, which requires extra copies of the string (and definitely has an impact when sending large messages.
Also since To_String needs to store the result string somewhere, it will be on the stack. This fails if the string is larger than 2Mb (with the default stack size for tasks)

briot:
I added a pragma Compile_Time_Error to ensure that Character'Size is the same as Stream_Element'Size, as you recommended a long time ago (sorry for the delay)

TurboGit (approved):
Works for me ! Thanks.

PR #161, opened 2020-06-25
Add missing calls to On_Close

briot:
When the websocket is closed on error, we should still call On_Close
 so that user-defined cleanup is run.
(Send): Avoid a case of double-free


TurboGit:
Looks good to me, thanks Manu !

PR #162, opened 2020-07-10
Config.Ini.Read: review documentation and fix implemenation.

TurboGit:
All exceptions where masked which was not compliant with the comments.

Now this routine properly propagates the exception Constraint_Error in case
of erronous option name or value.

The regression was introduced in 05a2aea when fixing a possible resource
leak failing to close the file descriptor of the configuration file in
case or error.

For T707-002.

### DIFF
--- a/src/core/aws-config-ini.adb
+++ b/src/core/aws-config-ini.adb
@@ -1,7 +1,7 @@
 ------------------------------------------------------------------------------
 --                              Ada Web Server                              --
 --                                                                          --
---                     Copyright (C) 2000-2019, AdaCore                     --
+--                     Copyright (C) 2000-2020, AdaCore                     --
 --                                                                          --
 --  This library is free software;  you can redistribute it and/or modify   --
 --  it under terms of the  GNU General Public License  as published by the  --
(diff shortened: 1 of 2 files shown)


TurboGit:
> and raise should be executed out of "if" block.

Indeed ! Good catch, thanks.

TurboGit:
> How about to catch all exceptions the same way, with raise at the end ?

Because this has been done in 05a2aea  on purpose. We don't want to raise exceptions except for error in ini file.

anisimkov:
Why should we hide absent file exception ? If user put wrong file name into call he do not know about error.

TurboGit:
Because a user may have a Read(C, "somefile") in their code that will actually read the file if present and do nothing otherwise. It is also not uncommon to change a config file to remove the read permissions just to avoid the server to use it. This is common Linux usage for daemons. So I don't think we should change this here, or at least it should be discussed in another ticket.

anisimkov:
Pascal, you removed notes
-- Raises Ada.Text_IO.Name_Error if Filename does
--  not exist.

I think it is wrong, we hafe Read_If_Present routine for this kind of behavior.

TurboGit:
> I think it is wrong, we have Read_If_Present routine for this kind of behavior.

No, this is internal and cannot be used by end-users. And again this change was done for 05a2aea. This is not related to this ticket or I fail to understand what are your goal!

anisimkov:
It is related to this TN, because you are removing notes from comment. Customer told that comment is not related to behaviour. I think that wrong not a comment, but behaviour, And we have to fix behaviour and remain comment unchanged.

anisimkov:
I propose anothe fix instead https://github.com/AdaCore/aws/pull/163

TurboGit:
Closing, see #163 .

PR #163, opened 2020-07-13
Restore exceptions logic on read configuration

anisimkov:
S223-004 T707-002

Ignore all exceptions on reading configuration at elaboration code.
Restore exceptions  propagation when load configuration by user.

### DIFF
--- a/src/core/aws-config.adb
+++ b/src/core/aws-config.adb
@@ -28,11 +28,11 @@
 ------------------------------------------------------------------------------
 
 with Ada.Characters.Handling;
-with Ada.Directories;
 with Ada.Environment_Variables;
 
 with AWS.Config.Ini;
 with AWS.OS_Lib;
+with AWS.Utils;
 
 package body AWS.Config is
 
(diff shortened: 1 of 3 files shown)


src/core/aws-config.adb
```diff
@@ -583,16 +582,17 @@ package body AWS.Config is
...
-   -- Read_If_Present --
-   ---------------------
+   --------------------
+   -- Read_Or_Ignore --
+   --------------------
 
-   procedure Read_If_Present (Filename : String) is
+   procedure Read_Or_Ignore (Filename : String) is
    begin
-      if Directories.Exists (Filename) then
```
TurboGit, line 592: I don't agree with this change as it uses exception as control flow.

TurboGit: We still need to check for Directory.Exists() and catch all other exceptions if needed.

TurboGit: Apart from that I'm ok with this change. We hide internal exception when reading ini file but still raise exception for explicit call to Read() from user code. Seem a good compromise.

anisimkov: Ok. I'll restore check file existence inside of Read_Or_Ignore.

PR #170, opened 2020-09-21
Improve sockets poll processing

anisimkov:
T521-006

Call poll again on EINTR even if timeout is over.


TurboGit (approved):
Looks fine, thanks!

PR #174, opened 2020-09-27
Fix WSDL duration image used in SOAP streaming.

TurboGit:
Continued work for T915-027.


anisimkov (approved):

anisimkov:
I would avoid to use Months and Years in duration format because back conversion can give different result. How long is one month 28 or 31 day ? I think it is possible to use only Days even if duration more than one month.

PR #178, opened 2020-10-13
Support wildcard patterns in Net.SSL.Add_Host_Certificate

OCTAGRAM:


anisimkov (commented):
It would be good to have a test for this feature.

anisimkov (approved):

anisimkov (changes requested):
This modification should be done with GNUTLS support and with testsuite. I'll do it soon.

PR #179, opened 2020-10-14
fix zlib stream on empty files

TurboGit:


anisimkov (commented):
I don't think it is right fix. The next elsif condition is ignoring the Z_BUF_ERROR for some reason. I need time to understand.

TurboGit:
> I don't think it is right fix. The next elsif condition is ignoring the Z_BUF_ERROR for some reason. I need time to understand.

I've seen that but it is hard to believe that a Z_BUF_ERROR is not reported as an error at this point. Anyway, let me know if you remember this part. TIA.

anisimkov:
See zlib.h <<Note that Z_BUF_ERROR is not fatal, and
  deflate() can be called again with more input and more output space to
  continue compressing.>>

PR #180, opened 2020-10-16
Detect error when input data to inflate is zero length

anisimkov:
TA13-028

Do not ignore Z_BUF_ERROR if no data on input.


TurboGit (approved):
Looks good, thanks !

PR #181, opened 2020-10-17
po/empty zlib

TurboGit:


anisimkov (changes requested):
There has to be expected exception.
Aaa. I see, the output is NOK: ....
Why is NOK ? Why not just OK ?

TurboGit:
Ok, I'll use OK.

anisimkov (approved):

PR #184, opened 2020-12-02
Fix Zlib end-of-stream detection.

TurboGit:
Add corresponding regression test.

For TC01-031.

### DIFF
--- a/include/zlib.adb
+++ b/include/zlib.adb
@@ -1,7 +1,7 @@
 ----------------------------------------------------------------
 --  ZLib for Ada thick binding.                               --
 --                                                            --
---  Copyright (C) 2002-2019, Dmitriy Anisimkov                --
+--  Copyright (C) 2002-2020, Dmitriy Anisimkov                --
 --                                                            --
 --  Open source license information is in the zlib.ads file.  --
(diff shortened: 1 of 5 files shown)


anisimkov (commented):
commit message told that it is the test, but I see the change in zlib.adb too.

TurboGit:
No the first line is "Fix Zlib end-of-stream detection." and it the subject of this PR.

anisimkov (commented):
Could you add some explanation of the problem to commit message.
As I can read from test, the failure does not happen when we remove few last bytes from compressed file. As I can understand it is bug too.

TurboGit:
New version proposed.

I think this is correct and I don't see how some bytes could remain on the stream since the Read() on the stream did return with Last = 0 (no more data read). Also there is no regression. For sure you know this part of AWS better than me so if you have a better proposal let me know.

anisimkov (commented):
I think need to add condition that there was no data in output
(No_Byte and then Last < Item_First and then V_Flush = Finish)

TurboGit:
New version pushed, it is working using your proposal. Thanks.

anisimkov (approved):

PR #185, opened 2020-12-05
More common end-of-stream detection

anisimkov:
TC01-031

Raise exception when there was no Z_STREAM_END return code and there is no
input data and there is no output data and Flush = Finish. This way the
error will be detected not only in Read but also in Write and
Generic_Translate routines.
Test improved to detect end-of-stream error even if only one byte removed
from the end of file.


TurboGit (approved):
Looks ok, thanks !

PR #194, opened 2021-01-28
Fix issue where From_Utf8 was raising an exception.

TurboGit:
Some characters don't have an equivalence from Utf8 to 8bits. In
this case we fallback by default to a space character to avoid
the exception. A callback is also put in place to let user's propose
some equivalence better the Utf8 character and 8bit one. For example
in Utf8 there is different left and right apostrophe, one can
use the standard 8bit quote in place.

For U127-049.

### DIFF
--- a/src/soap/soap-utils.adb
+++ b/src/soap/soap-utils.adb
@@ -1,7 +1,7 @@
 ------------------------------------------------------------------------------
 --                              Ada Web Server                              --
 --                                                                          --
---                     Copyright (C) 2000-2020, AdaCore                     --
+--                     Copyright (C) 2000-2021, AdaCore                     --
 --                                                                          --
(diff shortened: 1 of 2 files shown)


anisimkov (commented):
Why is it space, Maybe the question mark is better ? I saw the '?' when unable to decode wide char somewhere,

TurboGit:
> Why is it space, Maybe the question mark is better ? I saw the '?' when unable to decode wide char somewhere,

Good point indeed, probably better will do that. Thanks.

anisimkov (changes requested):
comment next to Set_Utf8_Mapping

TurboGit:
> comment next to Set_Utf8_Mapping
> This routine can be useD

Fixed, thanks!

anisimkov (approved):

PR #197, opened 2021-02-12
Regenerate longer certificates

anisimkov:
no-tn-check

OpenSSL 1.1.1f claims no less then 2048 bit certificates.


TurboGit (approved):
Thanks

PR #202, opened 2021-03-24
Support content ranges larger than 2 GB

Kensan:
Use Ada.Streams.Stream_Element_Offset as base for
AWS.Client.Content_Bound type to increase addressable content ranges.


TurboGit (approved):
Looks good to me, thanks !

PR #204, opened 2021-03-29
Fix URL decoding.

TurboGit:
Remove the translation from plus sign (+) to space. This prevents
downloading any file having a + in its name.

Update the 0023_param2 regression test expected output which was actually
wrong as the name of the parameter was changed from "char+" to "char "
when decoding.

Fixes U326-027.

### DIFF
--- a/src/core/aws-url-set.adb
+++ b/src/core/aws-url-set.adb
@@ -1,7 +1,7 @@
 ------------------------------------------------------------------------------
 --                              Ada Web Server                              --
 --                                                                          --
---                     Copyright (C) 2007-2019, AdaCore                     --
+--                     Copyright (C) 2007-2021, AdaCore                     --
 --                                                                          --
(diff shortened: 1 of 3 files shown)


anisimkov (changes requested):
I think it is not right fix. The 0023_param2 do
Mist.Add ("char+", "value%23", Decode => True);
it mean that name and value decoded before put into container and "+" is coding space in HTTP parameters.

anisimkov:
I read some internet and made some experiments. The browsers use + to encode form parameters, but do not use + to encode pathnames. So + decoded into space only in query part, after first ? character in URL.

anisimkov:
<html><body><form>
<input name="inp name" value="some value">
<input type="submit" value="Submit">
</form></body></html>

TurboGit:
Good point indeed, new version should be better.

anisimkov (approved):

PR #206, opened 2021-04-04
Wildcard pattern in certificate support

anisimkov:
No need Indefinite_Vectors, Vectors is enough.


TurboGit (approved):
Looks good to me. Thanks.

PR #208, opened 2021-05-03
Task safe random generation

anisimkov:
U503-007

### DIFF
--- a/regtests/0336_random_task_safe/test.adb
+++ b/regtests/0336_random_task_safe/test.adb
@@ -1,0 +1,119 @@
+with Ada.Containers.Hashed_Sets;
+
+with Ada.Exceptions; use Ada.Exceptions;
+with Ada.Text_IO;    use Ada.Text_IO;
+with AWS.Utils;      use AWS.Utils;
+
+procedure Test is
+
+   type Test_Random_Mode is (None, Save, Lookup);
+
+   package Random_Sets is new Ada.Containers.Hashed_Sets
+     (Random_Integer, Ada.Containers.Hash_Type'Mod, "=");
+
+   Set : Random_Sets.Set;
+   Locker : RW_Semaphore (1);
+   Wrong_Data_Set : exception;
+
+   procedure Test_Random_Sequence
+     (Count : Positive; Mode : Test_Random_Mode);
+
+   function Test_Random_Sequence
+     (Initialor : Integer; Count : Positive) return String;
+
+   task Secondary is
+      entry Start (Count : Positive; Mode : Test_Random_Mode);
+      entry Done (Quit : Boolean);
+   end Secondary;
+
+   --------------------------
+   -- Test_Random_Sequence --
+   --------------------------
+
+   function Test_Random_Sequence
+     (Initialor : Integer; Count : Positive) return String is
+   begin
+      Random_Reset (Initialor);
+      Test_Random_Sequence (Count, None);
+      return Random_String (77);
+   end Test_Random_Sequence;
+
+   procedure Test_Random_Sequence
+     (Count : Positive; Mode : Test_Random_Mode)
+   is
+      R : Random_Integer;
+   begin
+      for J in 1 .. Count loop
+         R := Random;
+
+         case Mode is
+            when None =>
+               null;
+
+            when Save =>
+               Locker.Write;
+               Set.Include (R);
+               Locker.Release_Write;
+
+            when Lookup =>
+               if not Set.Contains (R) then
+                  raise Wrong_Data_Set with
+                    "Expected random element" & R'Img & " absent at" & J'Img;
+               end if;
+         end case;
+      end loop;
+
+   end Test_Random_Sequence;
+
+   ---------------
+   -- Secondary --
+   ---------------
+
+   task body Secondary is
+      Count : Positive;
+      Mode  : Test_Random_Mode;
+      Quit  : Boolean;
+   begin
+      loop
+         accept Start (Count : Positive; Mode  : Test_Random_Mode) do
+            Secondary.Count := Count;
+            Secondary.Mode  := Mode;
+         end Start;
+
+         Test_Random_Sequence (Count, Mode);
+
+         accept Done (Quit : Boolean) do
+            Secondary.Quit := Quit;
+         end Done;
+
+         exit when Quit;
+      end loop;
+
+   exception
+      when E : others =>
+         Put_Line ("Secondary task: " & Exception_Message (E));
+   end Secondary;
+
+begin
+   if Test_Random_Sequence (321, 256) /= Test_Random_Sequence (321, 256) then
+      Put_Line ("Random_Reset with same initialor error");
+   end if;
+
+   for Mode in Save .. Lookup loop
+      Random_Reset (112344);
+      Test_Random_Sequence (4096, Mode);
+   end loop;
+
+   Set.Clear;
+
+   for Mode in Save .. Lookup loop
+      Random_Reset (44552211);
+      Secondary.Start (8196, Mode);
+      Test_Random_Sequence (8196, Mode);
+      Secondary.Done (Mode = Lookup);
+   end loop;
+
+exception
+   when E : others =>
+      Put_Line ("Main task: " & Exception_Message (E));
+end Test;
--- a/src/core/aws-utils.adb
+++ b/src/core/aws-utils.adb
@@ -68,7 +68,13 @@ package body AWS.Utils is
    --  Returns True if the string pointed to by Str and terminating to Last
    --  is well-formed UTF-8.
 
-   Random_Generator : Integer_Random.Generator;
+   protected Shared_Random is
+      function Generate return Random_Integer;
+      procedure Reset;
+      procedure Reset (Seed : Integer);
+   private
+      Random_Generator : Integer_Random.Generator;
+   end Shared_Random;
 
    ---------------------
    -- Append_With_Sep --
@@ -885,8 +891,17 @@ package body AWS.Utils is
 
    function Random return Random_Integer is
    begin
-      return Integer_Random.Random (Random_Generator);
+      return Shared_Random.Generate;
    end Random;
+
+   ------------------
+   -- Random_Reset --
+   ------------------
+
+   procedure Random_Reset (Seed : Integer) is
+   begin
+      Shared_Random.Reset (Seed);
+   end Random_Reset;
 
    -------------------
    -- Random_String --
@@ -1002,6 +1017,37 @@ package body AWS.Utils is
       end Seize_Internal;
 
    end Semaphore;
+
+   -------------------
+   -- Shared_Random --
+   -------------------
+
+   protected body Shared_Random is
+
+      --------------
+      -- Generate --
+      --------------
+
+      function Generate return Random_Integer is
+      begin
+         return Integer_Random.Random (Random_Generator);
+      end Generate;
+
+      -----------
+      -- Reset --
+      -----------
+
+      procedure Reset is
+      begin
+         Integer_Random.Reset (Random_Generator);
+      end Reset;
+
+      procedure Reset (Seed : Integer) is
+      begin
+         Integer_Random.Reset (Random_Generator, Seed);
+      end Reset;
+
+   end Shared_Random;
 
    -----------------------
    -- Significant_Image --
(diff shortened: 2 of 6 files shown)


briot:
I am curious why the complex implementation with task specific attributes ? Generally, this is for efficiency reasons, but here this is used when initially connecting a websocket (so very rarely in the lifetime of the socket itself, and a standard protected object seems to be efficient enough for the purpose (maybe one optimization is to `Get` four random numbers in a single protected procedure, to set the header) ?

anisimkov:
Random used not only in WebSockets, but for the HTTP sessions and HTTP Digest authentication. I don't think it is mach more  complex then just protected object.

briot:
How about a generic package, to share those controlled-task-specific attributes with AWS.Net.SSL__openssl.adb ? If for some reason T is null in your Finalize, for instance, you will not call the previous handler which might have significant impact in some applications. Also, if TV.Finalizer raises an exception, you will not do proper cleanup either (so missing exception handler in Finalize I think).

But if this is for performance reasons, I think a simpler protected object would be equivalent, and the code is easier to read. Semantically, there is no reason to want a per-task generator, right ?

anisimkov:
> How about a generic package, to share those controlled-task-specific attributes with AWS.Net.SSL__openssl.adb ?
I don't understand this. I fill real overcomplexity here.

> If for some reason T is null in your Finalize, for instance, you will not call the previous handler which might have significant impact in some applications.

If TV = null it can only mean that Finalizer already was called because I set attribute together with Finalizer.
I just moved Set_Value before Task_Termination.Set_Specific_Handler to be sure.

> if TV.Finalizer raises an exception, you will not do proper cleanup either (so missing exception handler in Finalize I think).
The Finalizer only to avoid memory leak, so not very critical. If TV.Finalizer raise exception is incorrect behavior of prev handler. So I can only cleanup task attribute. It will be better.

> But if this is for performance reasons, I think a simpler protected object would be equivalent
I don't think that performance is equivalent, It is already used a much and we do not know the future using of thei random generator.

briot:
In my experience, and given that we are doing or are about to do I/O when this random number generator is called, the cost of the mutex for the protected type is dwarfed by the I/O. But you might have different measures.

anisimkov:
> In my experience, and given that we are doing or are about to do I/O 

OK, We decided to use protected object. I'll do the patch tomorrow.

src/core/aws-utils.adb
```diff
@@ -883,34 +887,64 @@ package body AWS.Utils is
...
+      return (if Task_Safe
+              then Random_Protecter.Random
+              else Integer_Random.Random (Random_Generator));
    end Random;
 
+   ----------------------
+   -- Random_Protecter --
+   ----------------------
+
+   protected body Random_Protecter is
```
TurboGit, line 901: I would rename this Shared_Random as we have Shared_Sock in AWS.

src/core/aws-utils.adb
```diff
@@ -883,34 +887,64 @@ package body AWS.Utils is
    -- Random --
    ------------
 
-   function Random return Random_Integer is
+   function Random (Task_Safe : Boolean := False) return Random_Integer is
```
TurboGit, line 890: Shouldn't we be safe by default ? I mean in AWS most calls are with Task_Safe True now no?

src/core/aws-utils.adb
```diff
@@ -883,34 +887,64 @@ package body AWS.Utils is
...
+         return Integer_Random.Random (Random_Generator);
+      end Random;
+
+   end Random_Protecter;
+
+   ------------------
+   -- Random_Reset --
+   ------------------
+
+   procedure Random_Reset (Initiator : Integer) is
```
TurboGit, line 918: Usually this is called a Seed for random numbers. (Initiator -> Seed).

src/core/aws-utils.ads
```diff
@@ -80,20 +80,33 @@ package AWS.Utils is
...
-   function Random_String (Length : Natural) return String with
+   function Random_String
+     (Length : Natural; Task_Safe : Boolean := False) return String with
      Inline,
      Post => Random_String'Result'Length = Length
              and then (for all C of Random_String'Result
                          => C in '0' .. '9' | 'a' .. 'z' | 'A' .. 'Z');
    --  Returns random string
 
+   procedure Random_Reset (Initiator : Integer);
```
TurboGit, line 104: Initiator -> Seed

src/core/aws-utils.ads
```diff
@@ -80,20 +80,33 @@ package AWS.Utils is
...
+   --  This function need only if user want to get predictable random numbers.
+   --  It mean that after call to Random_Reset with same Initialor the same
```
TurboGit, line 106: Initiator -> Seed

src/core/aws-utils.ads
```diff
@@ -80,20 +80,33 @@ package AWS.Utils is
...
 
-   function Random return Random_Integer;
-   --  Returns a random integer number
-
-   procedure Random_String (Item : out String) with
-     Post => (for all C of Item => C in '0' .. '9' | 'a' .. 'z' | 'A' .. 'Z');
+   function Random (Task_Safe : Boolean := False) return Random_Integer;
+   --  Returns a random integer number.
+   --  Task_Safe parameter need to avoid concurrency on modifying random
+   --  generator internal state. Note that Task_Safe = False does not mean
```
TurboGit, line 86: I see, so my comment about have safe by default is maybe not an issue.

TurboGit (changes requested):

src/core/aws-utils.adb
anisimkov: Done

src/core/aws-utils.adb
anisimkov: Changed to default True

src/core/aws-utils.adb
anisimkov: Done

src/core/aws-utils.ads
anisimkov: Done

src/core/aws-utils.ads
anisimkov: I read this after I moved Task_Safe default to True. Maybe we need global flag for this mode. If anybly want to detect races, he can set it to True.

briot: But it is not about "detecting races" just to satisfy a tool. It is about avoiding them in the first place. AWS doesn't have optional tasks, so the code needs to decide for every call whether there are multiple tasks involved or not. A global default doesn't really make sense here.

anisimkov: I created separate routine to Set_Random_Task_Safe and removed Task_Safe parameter from random routines. Entire commit bacame smaller.

anisimkov: <<It is about avoiding them in the first place.>> GNAT random generator is OK with concurrently update of  the internal state. If you need to satisfy the race detection tools you can use Set_Random_Task_Safe routine.

briot: I don't understand how you come to the conclusion that the random generator is task safe. AWS uses Ada.Numerics.Discrete_Random, for which the generator is from System.Random_Numbers. There are no protected types, nor atomic types, nor mutexes in that package, so I do not see how it would be safe to use from multiple tasks. Can you explain ?

anisimkov: it is not really task safe. Internal state mixed when modified concurrently, but it is not lead to failure in random numbers generation. The random numbers became a little bit more random.

briot: Actually no, they just become garbage. For instance, they no longer follow the ARM requirement that they be uniformly distributed. For all I know, we might end up with invalid values in the generator which means that from now on we are getting the same number always.
Anything that is not task safe must be protected in the context of AWS, which uses tasking heavily and on multiple platforms. Thread-sanitizer is an excellent tool to find such cases where things seem to work but only work by chance and could break when moving to new versions of the system, new platform, new compiler, or even new phase of the moon.

anisimkov: You are right, I just made experiment and I see that the concurrency lead to the repeating of the same numbers is sequence.
I'll chenge the implementation to always protected.

briot: Thanks Dmitriy !

src/core/aws-utils.ads
```diff
@@ -94,6 +101,11 @@ package AWS.Utils is
                          => C in '0' .. '9' | 'a' .. 'z' | 'A' .. 'Z');
    --  Returns random string
 
+   procedure Random_Reset (Seed : Integer);
+   --  This function need only if user want to get predictable random numbers.
```
TurboGit, line 105: is needed only if the user

src/core/aws-utils.ads
```diff
@@ -94,6 +101,11 @@ package AWS.Utils is
...
+   --  It mean that after call to Random_Reset with same Seed the same sequence
```
TurboGit, line 106: It means that after calling Random_Reset with the same seed the same sequence
of Random and Random_String calls will give the same results.

src/core/aws-utils.adb
```diff
@@ -70,6 +70,12 @@ package body AWS.Utils is
 
    Random_Generator : Integer_Random.Generator;
 
+   protected Shared_Random is
+      function Generate return Random_Integer;
+   end Shared_Random;
+
+   Random_Task_Safe : Boolean := False;
```
TurboGit, line 77: Should be safe by default. User should not care about this except if performance is an issue.

src/core/aws-utils.ads
```diff
@@ -80,6 +80,13 @@ package AWS.Utils is
    --  General helper functions --
    -------------------------------
 
+   procedure Set_Random_Task_Safe (Flag : Boolean := True);
```
TurboGit, line 83: In fact I would just remove this and have a unique and safe mode. As discussed the random generator is not called a lot and not in critical part of AWS.

TurboGit (changes requested):

src/core/aws-utils.ads
anisimkov: Yes, it is protected always now.

src/core/aws-utils.ads
anisimkov: done

src/core/aws-utils.ads
anisimkov: done

src/core/aws-utils.adb
anisimkov: this flag removed. We are protected always.

regtests/0336_random_task_safe/test.adb
```diff
@@ -0,0 +1,119 @@
...
+            when None =>
+               null;
+
+            when Save =>
+               Locker.Write;
+               Set.Include (R);
+               Locker.Release_Write;
+
+            when Lookup =>
+               if not Set.Contains (R) then
```
briot, line 59: shouldn't we call `Locker.write` before we read from the set ?

src/core/aws-utils.ads
```diff
@@ -94,6 +94,12 @@ package AWS.Utils is
                          => C in '0' .. '9' | 'a' .. 'z' | 'A' .. 'Z');
    --  Returns random string
 
+   procedure Random_Reset (Seed : Integer);
+   --  This function is needed only if the user want to get predictable random
```
briot, line 98: "want" -> "wants"

briot (dismissed):

regtests/0336_random_task_safe/test.adb
anisimkov: Nope, we are not reading when writing. First phase is writing concurrently, second phase reading concurrently. Concurrent reading does not require protection.

src/core/aws-utils.ads
anisimkov: Fixed, Thanks

regtests/0336_random_task_safe/test.adb
briot: A few years ago, reading from containers was actually also changing some internal flags (hence the `pragma Suppress (Container_Checks)` that was introduced at the time. I forgot whether it has been fixed since then.

anisimkov: When Lock and Busy counter became atomic incremented/decremented the cursor tempering check became RO task safe.

briot: thanks for confirming ! useful to know

TurboGit (approved):
Looks good to me, thanks !

regtests/0336_random_task_safe/test.adb
anisimkov: The only case when the RO operations with tempering check is not task safe, when platform does not support atomic increments. It this case we should suppress tempering check for RO concurrency available.

Issue #17, opened 2018-03-09
Problem building (debian 9 x64)


ohenley:
Hi, 
Indications, as how to build, are not working on a Debian 9 "stretch" fresh install. After investigating how to fix the build on my side, I let go as I am no expert in Makefiles. Here are the different outputs:

$ make setup build install
gprbind xoscons.bexch
gnatbind xoscons.ali
gnatgcc -c b__xoscons.adb
gnatgcc xoscons.o -o xoscons
Setup OS specific definitions
Can not generate system tags. The test are disabled
make[1]: *** No rule to make target 'setup'.  Stop.
Makefile:162: recipe for target 'templates_parser_setup' failed
make: *** [templates_parser_setup] Error 2

Then I tried just 

$ make
object directory "/home/ohenley/gnoga-code-1.2b/deps/aws/tools/../.build/x86_64-linux-gnu/release/static/obj/tools/" created
object directory "/home/ohenley/gnoga-code-1.2b/deps/aws/include/zlib/../../.build/x86_64-linux-gnu/release/static/obj/zlib" created for project zlib
library directory "/home/ohenley/gnoga-code-1.2b/deps/aws/include/zlib/../../.build/x86_64-linux-gnu/release/static/lib/zlib" created for project zlib
object directory "/home/ohenley/gnoga-code-1.2b/deps/aws/include/../.build/x86_64-linux-gnu/release/static/obj/include" created for project include
library directory "/home/ohenley/gnoga-code-1.2b/deps/aws/include/../.build/x86_64-linux-gnu/release/static/lib/include" created for project include
object directory "/home/ohenley/gnoga-code-1.2b/deps/aws/ssl/../.build/x86_64-linux-gnu/release/static/obj/ssl" created for project ssl
library directory "/home/ohenley/gnoga-code-1.2b/deps/aws/ssl/../.build/x86_64-linux-gnu/release/static/lib/ssl" created for project ssl
object directory "/home/ohenley/gnoga-code-1.2b/deps/aws/src/../.build/x86_64-linux-gnu/release/static/obj/src" created for project src
library directory "/home/ohenley/gnoga-code-1.2b/deps/aws/src/../.build/x86_64-linux-gnu/release/static/lib/src" created for project src
library directory "/home/ohenley/gnoga-code-1.2b/deps/aws/.build/x86_64-linux-gnu/release/static/lib/aws" created for project aws
[...]

Issue #16, opened 2018-02-28
4 issues related to AWS.Client proxy support


OCTAGRAM:
AWS.Client has broken proxy support. Mostly in AWS.Client.HTTP_Utils:

First, port must be sent unconditionally, or else there are issues with some proxy software. "CONNECT host:443 HTTP/1.1" as opposed to "CONNECT host HTTP/1.1".

On line 147 replace
`& Port_Not_Default (Connection.Host_URL);`
with
`& ':' & Port (Connection.Host_URL);`

Second, SSL configuration is not being taken into consideration when connection goes through proxy. It makes use of default configuration instead, and default configuration demands unconditionally client certificate which is missing of course, thus any secure connection through proxy fails due to missing cert.pem.

On line 209 replace
`SS : Net.SSL.Socket_Type := Net.SSL.Secure_Client (Sock.all);`
with
`SS : Net.SSL.Socket_Type := Net.SSL.Secure_Client (Sock.all, Connection.SSL_Config);`

(style checks require dividing it into more lines)

Third, SNI is not being sent via proxy. AWS.Net.SSL.Connect sends SNI, but AWS.Net.SSL.Secure_Client does not, and sending SNI is not being exported publicly. Secure_Client should probably be extended with SNI host argument and perform SNI. Maybe Connect code should be moved to Secure_Client, and Connect must invoke Secure_Client to make sure behavior is the same no matter if there are proxies or not.

Forth, SSL session reuse is not implemented for proxies. I do not quite understand this thing. When proxy is being used, it can only be desirable to share sessions going through the same proxy, but deny sharing sessions between different proxies. So maybe disabled session sharing is a good thing.

anisimkov:
I comitted 2 first propsals. For the rest need more investigations.

anisimkov:
3rd issue is fixed

anisimkov:
<<it can only be desirable to share sessions going through the same proxy>>
I think we don't have limit to use same SSL session over different proxies.
But anyway HTTP_Connection is tired to proxy and we can't change it after connection created.

I made a fix but can't check it. If you defined that the session was not reused, maybe you can confirm that it is fixed ?

OCTAGRAM:
Can say little about 4th one, just noted the asymmetry and mentioned it for completeness. Cannot confirm this part.

I had impression that there is some kind of pool like in browser, but seems no.

anisimkov:
Ok. Thank you for the analysis anyway.
[...]

Issue #12, opened 2017-12-02
problems with making documentation


evanescente-ondine:
=> make build-doc          
=== Build doc
gcc -c -I../src/core -I../src/soap -O0 -o /home/mehdi/essai/aws/docs/build/obj/build.o build.adb
gcc -c -I./ -I../src/core -I../src/soap -O0 -I- -o /home/mehdi/essai/aws/docs/build/obj/aws.o /home/mehdi/essai/aws/src/core/aws.ads
gcc -c -I./ -I../src/core -I../src/soap -O0 -I- -o /home/mehdi/essai/aws/docs/build/obj/aws-default.o /home/mehdi/essai/aws/src/core/aws-default.ads
gcc -c -I./ -I../src/core -I../src/soap -O0 -I- -o /home/mehdi/essai/aws/docs/build/obj/soap.o /home/mehdi/essai/aws/src/soap/soap.ads
gnatbind -aO/home/mehdi/essai/aws/docs/build/obj -I../src/core -I../src/soap -x /home/mehdi/essai/aws/docs/build/obj/build.ali
gnatlink /home/mehdi/essai/aws/docs/build/obj/build.ali -o build/bin/build
make[1]: sphinx-build: Command not found
Makefile:72: recipe for target 'html' failed
make[1]: *** [html] Error 127
Makefile:69: recipe for target 'build-doc' failed
make: *** [build-doc] Error 2

TurboGit:
Yes, as reported in the log above, you're missing the sphinx-build command which comes with the python-sphinx on GNU/Debian.

Issue #10, opened 2017-08-28
link error AWS with SSL on debian 9 64 bit


b201b201:
Hi!
The following problem was posted on AWS-mailing list 2017-08-20 and
did not receive any reply.

Having a production machine on 64 bit debian 8 I got

bnl@prod:~$ apt-cache pkgnames | grep -i libssl | sort
libssl1.0.0
libssl1.0.0-dbg
libssl-dev
libssl-doc
libssl-ocaml
libssl-ocaml-dev

and a dev machine on 64 bit debian 9 with

bnl@tp:~/svn/bnlbot/botstart/bot-1-0/source/ada$ apt-cache pkgnames |
grep libssl | sort
libssl1.0.0
libssl1.0.2
libssl1.0-dev
libssl1.1
libssl-dev
libssl-doc
libssl-ocaml
libssl-ocaml-dev
bnl@tp:~/svn/bnlbot/botstart/bot-1-0/source/ada$

To build AWS with ssl support I set

SOCKET    = openssl

in makefile.conf

Building AWS itself works on both debians but using it does not work on
debian 9

   [link]         poll.adb
/usr/local/ada/aws/gpl2017_ssl/lib/aws/libaws.a(aws-net-ssl__openssl.o):
In function `aws__net__ssl__locking__initializeXnn':
aws-net-ssl__openssl.adb:(.text+0x6b0a): undefined reference to
`CRYPTO_set_id_callback'
aws-net-ssl__openssl.adb:(.text+0x6b14): undefined reference to
`CRYPTO_set_locking_callback'
aws-net-ssl__openssl.adb:(.text+0x6b1e): undefined reference to
`CRYPTO_set_dynlock_create_callback'
aws-net-ssl__openssl.adb:(.text+0x6b28): undefined reference to
`CRYPTO_set_dynlock_lock_callback'
/usr/local/ada/aws/gpl2017_ssl/lib/aws/libaws.a(aws-net-ssl__openssl.o):
In function `aws__net__ssl__ciphers':
aws-net-ssl__openssl.adb:(.text+0x7e76): undefined reference to
`SSLv23_method'
/usr/local/ada/aws/gpl2017_ssl/lib/aws/libaws.a(aws-net-ssl__openssl.o):
In function `aws__net__ssl__signature__3':
aws-net-ssl__openssl.adb:(.text+0x9248): undefined reference to
`EVP_MD_CTX_create'
aws-net-ssl__openssl.adb:(.text+0x931e): undefined reference to
[...]

PR #209, opened 2021-05-04
Fix WebSocket memory free to be thread safe.

TurboGit:
For U504-028.

### DIFF
--- a/src/core/aws-net-websocket-registry.adb
+++ b/src/core/aws-net-websocket-registry.adb
@@ -160,7 +160,7 @@ package body AWS.Net.WebSocket.Registry is
       entry Get_Socket (WebSocket : out Object_Class);
       --  Get a WebSocket having some data to be sent
 
-      procedure Get_Socket (Id : UID; WebSocket : out Object_Class);
+      entry Get_Socket (Id : UID; WebSocket : out Object_Class);
       --  Get a WebSocket Id or null if the WebSocket is not registerred
       --  anymore. Marks the socket as in use, so that it will not get
       --  freed until Release_Socket is called.
@@ -235,6 +235,7 @@ package body AWS.Net.WebSocket.Registry is
       Signal      : Boolean := False;    -- Transient signal, release Not_Emtpy
       S_Signal    : Boolean := False;    -- Shutdown is in progress
       New_Pending : Boolean := False;    -- New pending socket
+      New_State   : Boolean := False;    -- A sokcet has been released
       Count       : Natural := 0;        -- Not counting signaling socket
       Registered  : WebSocket_Map.Map;   -- Contains all the WebSocket ref
       Pending     : WebSocket_List.List; -- Pending messages to be sent
@@ -638,14 +639,14 @@ package body AWS.Net.WebSocket.Registry is
          --  record this socket to be freed as soon as it is released
          --  (Release_Socket) call.
 
+         if Registered.Contains (WebSocket.Id) then
+            Unregister (Registered (WebSocket.Id));
+         end if;
+
          if WebSocket.State.Sending then
             WebSocket.State.To_Free := True;
 
          else
-            if Registered.Contains (WebSocket.Id) then
-               Unregister (Registered (WebSocket.Id));
-            end if;
-
             Release_Memory (Object (WebSocket.all));
             Unchecked_Free (WebSocket);
          end if;
@@ -725,13 +726,29 @@ package body AWS.Net.WebSocket.Registry is
       -- Get_Socket --
       ----------------
 
-      procedure Get_Socket (Id : UID; WebSocket : out Object_Class) is
+      entry Get_Socket (Id : UID; WebSocket : out Object_Class)
+        when New_State or else S_Signal
+      is
          C : constant WebSocket_Map.Cursor := Registered.Find (Id);
       begin
+         New_State := False;
+
          if WebSocket_Map.Has_Element (C) then
-            WebSocket := Registered (C);
-            WebSocket.State.Sending := True;
+            declare
+               W : constant Object_Class := Registered (C);
+            begin
+               if W.State.Sending then
+                  requeue Get_Socket;
+
+               else
+                  WebSocket := W;
+                  WebSocket.State.Sending := True;
+               end if;
+            end;
+
          else
+            --  The socket is not registerred anymore, just leave now
+
             WebSocket := null;
          end if;
       end Get_Socket;
@@ -837,6 +854,7 @@ package body AWS.Net.WebSocket.Registry is
 
          Registered.Insert (WebSocket.Id, WebSocket);
          Success := True;
+         New_State := True;
       end Register;
 
       --------------------
@@ -857,6 +875,8 @@ package body AWS.Net.WebSocket.Registry is
          else
             New_Pending := True;
          end if;
+
+         New_State := True;
       end Release_Socket;
 
       ------------
@@ -1204,6 +1224,7 @@ package body AWS.Net.WebSocket.Registry is
       begin
          Registered.Exclude (WebSocket.Id);
          WebSocket.State.Sending := False;
+         New_State := True;
 
          Remove (WebSocket);
          Signal_Socket;
@@ -1416,9 +1437,12 @@ package body AWS.Net.WebSocket.Registry is
          begin
             Send (WS, Message);
             WS.Messages.Delete_First;
-
-            DB.Release_Socket (WS);
+         exception
+            when others =>
+               null;
          end;
+
+         DB.Release_Socket (WS);
       end loop;
    end Message_Sender;
 


src/core/aws-net-websocket-registry.adb
```diff
@@ -161,10 +161,14 @@ package body AWS.Net.WebSocket.Registry is
       entry Get_Socket (WebSocket : out Object_Class);
       --  Get a WebSocket having some data to be sent
 
-      procedure Release_Socket (WebSocket : Object_Class);
+      procedure Release_Socket (WebSocket : in out Object_Class);
       --  Release a socket retrieved with Get_Socket above, this socket will be
       --  then available again.
 
+      procedure Free (WebSocket : in out Object_Class);
+      --  Free WebSocket immediately if not taken by another tasK, otherwise
```
briot, line 169: "tasK" -> "task"

src/core/aws-net-websocket-registry.adb
```diff
@@ -568,6 +572,24 @@ package body AWS.Net.WebSocket.Registry is
...
+         if Sending.Contains (WebSocket.Id) then
+            WebSocket.To_Free := True;
+         else
+            Unchecked_Free (WebSocket);
+         end if;
+      end Free;
+
       ----------------
       -- Get_Socket --
       ----------------
```
briot, line 595: Shouldn't Get_Socket also avoid sockets that are marked with To_Free=True ?

src/core/aws-net-websocket-registry.adb
```diff
@@ -568,6 +572,24 @@ package body AWS.Net.WebSocket.Registry is
...
+      ----------
+
+      procedure Free (WebSocket : in out Object_Class) is
+      begin
+         --  If WebSocket is in Sending it means that it has been
+         --  taken by the Get_Socket call. We cannot free it now, we
+         --  record this socket to be freed as soon as it is released
+         --  (Release_Socket) call.
+
+         if Sending.Contains (WebSocket.Id) then
```
briot, line 586: I was looking in the code why Sending is a set. It seems to be it might just as well be a flag in the Object itself, rather than an independent set. The difficulty is to make sure it is only accessed via the protected object (though it could also be made atomic for instance).

src/core/aws-net-websocket-registry.adb
briot: There is a confusion in the API because the websocket itself has a primitive operation Free (aws-net-websocket.adb:215) which does an immediate deallocation. Shouldn't the latter internally call `Release_Socket` instead, and perform the actual deallocation later as needed ? As it is, the user might still free the socket directly, breaking the code. Perhaps you have an idea how to make this clearer ?
At the very least, it seems we should document that behavior, and perhaps rename AWS.Net.WebSocket.Registry.DB.Free to something like `Deferred_Free` or something to clear possible confusion in the AWS code itself ?

briot (commented):
I will try to create a test for the wrong behavior, or at least test on my own server.

briot:
Also noticed aws-net-websocket-registry.adb:445 which does a direct call to Unchecked_Free on the socket. Should this one be a call to Release_Socket instead ?

briot:
(in my own patch, I had replaced Unchecked_Free with a call to `AWS.Net.WebSocket.Free (WebSocket);`, but it was a long time ago and I forgot why it was needed

briot:
Similar call in `Close` and `On_Close`, it does Unchecked_Free directly, not clear to be whether it should be Release_Socket instead. I had also replaced this one with AWS.Net.WebSocket.Free in my patch.

briot:
At first sight, the use-after-free issue seems gone with this patch. I have two more reports from thread-sanitizer, but they do not seem to be the exact same issue

briot:
I ran my test multiple times with this patch, and did not notice the same thread-sanitizer report on use-after-free. There are however a few data races (two in the context of AWS, and one already reported to AdaCore in the runtime), which makes things a bit harder to read. I'll need some time to analyze.
At first sight the patch looks good

briot:
Unfortunately I think I just had the same issue with thread-sanitizer:
```
WARNING: ThreadSanitizer: heap-use-after-free
  Read of size 8 at 0x7b60000c6010 by thread T30:
    #0 aws__net__websocket__registry__message_reader__read_message aws-net-websocket.adb:469
    #1 aws__net__websocket__registry__message_readerTB aws-net-websocket-registry.adb:367
    #2 system__tasking__stages__task_wrapper s-tassta.adb:1201

  Previous write of size 8 at 0x7b60000c6010 by thread T15 (mutexes: write M457):
    #0 free tsan_interceptors_posix.cpp:707
    #1 __gnat_free s-memory.adb:122
    #2 system__pool_global__deallocate s-pooglo.adb:134
    #3 system__storage_pools__subpools__deallocate_any_controlled s-stposu.adb:470
    #4 aws__net__websocket__free__P149b__2.90 <null>
    #5 aws__net__websocket__free__2 aws-net-websocket.adb:199
    #6 aws__net__websocket__registry__db__closeN__2 aws-net-websocket-registry.adb:491
    #7 aws__net__websocket__registry__db__closeP__2 aws-net-websocket-registry.adb:469
    #8 aws__net__websocket__registry__close__3 aws-net-websocket-registry.adb:1150
    #9 gui__senders__gs_send_ws__process_message__do_send gui-senders.adb:168
    #10 gnatcoll__strings__access_string__dispatch gnatcoll-strings_impl.adb:694
    #11 gnatcoll__strings__access_string__do_access gnatcoll-strings_impl.adb:673
    #12 gnatcoll__strings__access_string gnatcoll-strings_impl.adb:699
    #13 gui__senders__gs_send_ws__process_message gui-senders.adb:178
    #14 gui__senders__gs_send_wsTB gui-senders.adb:190
    #15 system__tasking__stages__task_wrapper s-tassta.adb:1201
```

I double-checked that the patch is correctly applied. Since we are not using a protected type in message_reader, I think we can have this error at any time if it happens that we free the socket at the wrong time.

briot:
Here is a test similar to the one I am using. It shows thread-sanitizer reports for data races, but I haven't yet seen the same use-after-free error. I do recompile the AWS file (and the runtime files) with the same -fsanitize=thread switch, which might be needed to reproduce some of the errors.


```
with Ada.Unchecked_Deallocation;
with AWS.Config.Set;
with AWS.Net.WebSocket.Registry.Control;
with AWS.Server;
with AWS.Services.Dispatchers.URI;
with AWS.Status;
with Client;

procedure Main is

   procedure Unchecked_Free is new Ada.Unchecked_Deallocation
      (Client.Client_Type, Client.Client_Access);

   function Create_Socket
     (Request : AWS.Status.Data) return AWS.Net.WebSocket.Object_Class is
   begin
      return new AWS.Net.WebSocket.Object;  --  server-side socket
   end Create_Socket;

   Arr     : Client.Client_Array (1 .. 60);
   Conf    : AWS.Config.Object := AWS.Config.Get_Current;
   Handler : AWS.Services.Dispatchers.URI.Handler;
begin
   AWS.Config.Set.Reuse_Address (Conf, True);
   AWS.Net.WebSocket.Registry.Register
      ("/streams", Create_Socket'Unrestricted_Access);
   AWS.Server.Start
      (Web_Server => Client.Server,
       Config     => Conf,
       Dispatcher => Handler);
   AWS.Net.WebSocket.Registry.Control.Start;

   for A in Arr'Range loop
      Arr (A) := new Client.Client_Type (Id => A);
   end loop;

   delay 20.0;

   Client.Finish := True;

   for A in Arr'Range loop
      Arr (A).Terminated;

      --  Uncomment this to get thread-sanitizer error in runtime
--      Unchecked_Free (Arr (A));
   end loop;

   AWS.Net.WebSocket.Registry.Control.Shutdown;
   AWS.Server.Shutdown (Client.Server);
end Main;


with AWS.Server;
package Client is
   task type Client_Type (Id : Positive) is
      entry Terminated;
   end Client_Type;

   type Client_Access is access Client_Type;
   type Client_Array is array (Natural range <>) of Client_Access;

   Finish : Boolean := False with Atomic;
   --  Request termination of tasks

   Server : AWS.Server.HTTP;

end Client;



with Ada.Exceptions;
with Ada.Numerics.Float_Random;
with Ada.Text_IO;
with AWS.Net.WebSocket;
with AWS.Server.Status;
package body Client is

   task body Client_Type is
      use AWS.Net.WebSocket;
      Sock : AWS.Net.WebSocket.Object_Class;
      Gen  : Ada.Numerics.Float_Random.Generator;
   begin
      Ada.Numerics.Float_Random.Reset (Gen);
      while not Finish loop
         Sock := new AWS.Net.WebSocket.Object;  --  client-side socket
         AWS.Net.WebSocket.Connect
            (Sock.all,
             AWS.Server.Status.Local_URL (Server) & "/streams");

         Sock.Send ("some message");
         delay 3.0 * Duration (Ada.Numerics.Float_Random.Random (Gen));

         Sock.Close ("no longer needed");
         --  AWS.Net.WebSocket.Free (Sock);
      end loop;

      accept Terminated;

   exception
      when E : others =>
         Ada.Text_IO.Put_Line (Ada.Exceptions.Exception_Information (E));
         accept Terminated;
   end Client_Type;
end Client;



with "aws.gpr";
project Default is
   for Main use ("main.adb");

   package Compiler is
      for Switches ("Ada") use
         ("-O2", "-fsanitize=thread", "-fno-omit-frame-pointer");
   end Compiler;

   package Linker is
      for Switches ("Ada") use ("-fsanitize=thread", "-fuse-ld=gold");
   end Linker;
end Default;

**Review by anisimkov — CHANGES_REQUESTED** [review]
Looks like a typo in commit message "to ne thread"

**Comment by briot** [comment]
I'd like your input on the testcase above: you'll notice is calls the primitive `Close` directly on the socket. Should it instead call `AWS.Net.WebSocket.Registry.Close` ?
Also, I have commented out the call to Unchecked_Free for the socket, because it seems like user code should not free memory directly, but I am not sure how to free memory then.

File: src/core/aws-net-websocket-registry.adb (hunk 2 above)
Comment (TurboGit) [reply src/core/aws-net-websocket-registry.adb:595]: No because we do return only socket not in Sending container and in this case if it was to be free it would have been already done in Release_Socket.

File: src/core/aws-net-websocket-registry.adb (hunk 3 above)
Comment (TurboGit) [reply src/core/aws-net-websocket-registry.adb:586]: Yes, maybe but could be done as a clean-up and does not seem do be the issue at hand.

**Comment by TurboGit** [comment]
> Looks like a typo in commit message "to ne thread"

Yep, now fixed.

**Comment by briot** [comment]
A slightly different take on the take thread-sanitizer report I sent. This one doesn't involve free, but still shows a lack of mutex somewhere. One of the tasks in my server is writing to the websocket. This apparently doesn't involve any mutex (maybe my code is responsible for getting one, but this isn't clear from the API what such a mutex/protected object would be). Later, the same socket is written to by message_reader also without a mutex. So it is possible that both write would happen at the same time, leading to garbage of course.
```
WARNING: ThreadSanitizer: data race
  Write of size 1 at 0x7b7400501e10 by thread T30:
    #0 memmove sanitizer_common_interceptors.inc:789
    #1 memmove sanitizer_common_interceptors.inc:787
    #2 aws__net__buffered__write aws-net-buffered.adb:448
    #3 aws__net__websocket__protocol__rfc6455__send_frame_header aws-net-websocket-protocol-rfc6455.adb:767
    #4 aws__net__websocket__protocol__rfc6455__send_frame aws-net-websocket-protocol-rfc6455.adb:677
    #5 aws__net__websocket__protocol__rfc6455__receive aws-net-websocket-protocol-rfc6455.adb:501
    #6 aws__net__websocket__receive__2 aws-net-websocket.adb:549
    #7 aws__net__websocket__registry__message_reader__read_message aws-net-websocket.adb:469
    #8 aws__net__websocket__registry__message_readerTB aws-net-websocket-registry.adb:367
    #9 system__tasking__stages__task_wrapper s-tassta.adb:1201

  Previous read of size 8 at 0x7b7400501e10 by thread T17:
    #0 sendto sanitizer_common_interceptors.inc:6446
    #1 gnat__sockets__thin__c_sendto g-socthi.adb:362
    #2 gnat__sockets__send_socket g-socket.adb:2458
    #3 aws__net__std__send aws-net-std__gnat.adb:682
    #4 aws__net__websocket__send__5 aws-net-websocket.adb:574
    #5 aws__net__send aws-net.adb:474
    #6 aws__net__buffered__flush aws-net-buffered.adb:75
    #7 aws__net__websocket__protocol__rfc6455__send aws-net-websocket-protocol-rfc6455.adb:652
    #8 aws__net__websocket__send__4 aws-net-websocket.adb:612
    #9 aws__net__websocket__send__2 aws-net-websocket.adb:590
    #10 gui__senders__send gui-senders-send.adb:18
    #11 gui__senders__gs_send_ws__process_message__do_send gui-senders.adb:155
    #12 gnatcoll__strings__access_string__dispatch gnatcoll-strings_impl.adb:694
    #13 gnatcoll__strings__access_string__do_access gnatcoll-strings_impl.adb:673
    #14 gnatcoll__strings__access_string gnatcoll-strings_impl.adb:699
    #15 gui__senders__gs_send_ws__process_message gui-senders.adb:178
    #16 gui__senders__gs_send_wsTB gui-senders.adb:190
    #17 system__tasking__stages__task_wrapper s-tassta.adb:1201

```

File: src/core/aws-net-websocket-registry.adb (hunk 1 above)
Comment (TurboGit) [reply src/core/aws-net-websocket-registry.adb]: I thought about the `Deferred_Free` naming, not bad but confusing in the other direction as the implementation does free immediately if the socket is not currently handled.

Comment (briot) [reply src/core/aws-net-websocket-registry.adb]: `Free_Or_Defer` ?

**Comment by TurboGit** [comment]
Not sure if it will help but I have a new version of this PR. All Unchecked_Free in the registry are now done in the DB.Free() and so only possibly a deferred free.

One question, do you confirm that if the comment out the Unchecked_Free in the registry there is no crash ? At least this will make it clear that the only issue on the AWS WebSocket implementation is the WebSocket memory handling.

File: src/core/aws-net-websocket-registry.adb (hunk 1 above)
Comment (TurboGit) [reply src/core/aws-net-websocket-registry.adb]: Works for me! I'll do the change. Thanks.

**Comment by briot** [comment]
- don't we still have the possible issue of users doing a free on their own (`AWS.Net.WebSocket.Free`), or is that also handed out to `DB.Free` ?
- I do not know about crashes. My server crashes every few days, but I do not have a systematic reproducer (which of course heavily hints at invalid memory usage and race conditions). The reports I did (which the reproducer I sent earlier today will also show) are by using the thread-sanitizer tool which detects potential issues, even if they happen to have no influence during the specific build. You have to run such tests multiple times, because sanitizer might detect different things depending on the timing of the tasks.
- I actually do not believe freeing is the only issue, as highlighted by my last sanitizer report https://github.com/AdaCore/aws/pull/209#issuecomment-832689951 which shows possible data races when writing to the socket. This one might not be fatal to an application, but might end up sending corrupted data.

**Comment by TurboGit** [comment]
>     * don't we still have the possible issue of users doing a free on their own (`AWS.Net.WebSocket.Free`), or is that also handed out to `DB.Free` ?

Yes, users should not call AWS.Net.WebSocket.Free directly. This is on the spec to be able to override it for SSL implementation (as described in the comment). But this routine is needed as the garbage collection is done by Finalize().

**Comment by briot** [comment]
Can't `Free` be in the private part, if it needs to be overridden by specific parts of AWS, but not called by users ?
Come to think of it, the subprogram you need to override is the one that receives an `Object`, but the one that uses an `Object_Class` can probably be hidden altogether in some private part of even body ? At least users would need to instantiate Unchecked_Deallocation on their own, which might raise some kind of warning in their head. It seems to me that only the one that takes the `Object_Class` is dangerous. Otherwise, we still have a valid pointer, even though the socket itself might have become unusable (but then it can be checked)

An alternative could be to have `Free` call `Free_Or_Defer` in general, unless some special flag is set on the socket before (from `Finalize` or other places that legitimately need to free things). This defers checks to run time as opposed to compile time, but might be safer nonetheless...

**Comment by TurboGit** [comment]
> An alternative could be to have `Free` call `Free_Or_Defer` in general,

I thought about this, but how can it be possible. Free() call Free_Or_Defer() which does the Unchecked_Free() which calls the Free() because the object is finalized. Or maybe you meant something else ?

**Comment by briot** [comment]
I don't know precisely what I want, except it would be nice to have a safe API where people cannot easily shoot themselves in the foot by calling `AWS.Net.WebSocket.Free` (whose name is kind of tempting...)

I think we both agree that as soon as someone frees memory in an uncontrolled fashion, there is no way the registry and others will work correctly. In your sentence, there is ambiguity between the primitive `Free` and the one that applies to `Object_Class`. So let me rewrite it:

   `Free(Object_Class)` calls `Free_Or_Defer`, which does the `Unchecked_Free` (if possible), which calls `Free(Object)` because the object is finalized.

That seems to be working as expected, right ?

To me, the important is that Unchecked_Free only occurs when the websocket is no longer in use (either in the registry, or in the queue of messages read by `Messages_Reader`, or in user code). The latter is the major problem, because you do not know when the user is manipulating the websocket.

Maybe we need to somehow formalize the rules and ownership here...

* we hide `Free(Object_Class)`, users should not call this directly; Perhaps harder: remove Object_Class altogether in the public API (still needed internally of course), or use a type with access discriminant. Then users cannot free memory...
* Users should call `Close(Object_Class)` when they are done with the websocket. The effect is to close the actual C socket, and remove the socket from the registry. Perhaps also free all pending messages in the queue. `Close` needs to call the user's `On_Close`
* AWS itself might be calling `Close` on error, or when the client closes its end of the socket. In this case too, `On_Close` is called.
* After `On_Close` was called, users should not longer use the socket, which might get freed at any time convenient to AWS.
* Since the socket was unregistered, no more messages will be queued for it, so messages_reader will not try to manipulate it either.
* We need additional mutexes to protect socket operations. In my last sanitizer report, we saw that there could be several tasks writing to the socket concurrently. With this proposal, it might also be the case that we are closing the socket while the background messages_reader task is using the socket (hence the need to defer the actual freeing).

Going on a tangent here: perhaps the idea with the type-with-discriminant is nice: users can no longer store the websocket in their own code, but always need to go via the registry. We then provide a procedure like:
```
    procedure Use_Socket (Callback : not null access procedure (WebSocket: in out Type_With_Discriminant));
    --  or a generic procedure if we prefer
```
while the callback executes, AWS garantees that memory is not freed for the websocket. That simplifies lifetime management significantly since only AWS is responsible for it and user has no way around that. This is the same kind of API we use for refcounted types like GNATCOLL.Strings. This `Use_Socket` is also a good place to have the mutexes in place to prevent multiple concurrent writers on the socket..

This breaks the current API slightly, but it prevents very dangerous issues. The current API needs some improvements anyway as shown in Maxim's report in #138.

**Comment by TurboGit** [comment]
> That seems to be working as expected, right ?

This makes sense indeed. I'll work on this soon and propose a new patch where the release of WebSocket object is fully controlled by the registry and done only when the WebSocket is not used anymore.

**Comment by briot** [comment]
Don't you like my proposal with the `Use_Socket` procedure, and users only manipulate UID objects ? I think this is the best way to protect against concurrent calls to `Send` and `Free`.  Free is not the only issue here, we also need to address the issue with write.

**Comment by TurboGit** [comment]
> Don't you like my proposal with the `Use_Socket` procedure, and users only manipulate UID objects ? I think this is the best way to protect against concurrent calls to `Send` and `Free`. Free is not the only issue here, we also need to address the issue with write.

The issue is that will break the API ! You meant concurrent write as the other issue ? Because the Send and Free should be fixed by the WIP we've been discussing.

**Comment by briot** [comment]
yes, I realize it breaks the API. Let's see if we can fix the issues with the current API. But if we cannot, I would argue breaking the API doesn't matter as much as getting correct and safe software.

Do you have any idea on whether people use the websocket part of AWS ? Given Maxim's #138 issue, I think the API is currently hard to use in practice and that might have refrain people from using it.

I haven't seen how the current proposal fixes the missing mutexes around calls to Send (https://github.com/AdaCore/aws/pull/209#issuecomment-832689951) but maybe I missed that. I'll review the patch when you have a new version.

**Comment by TurboGit** [comment]
New version of the patch proposed. It also contains the atomic fix for Shutdown_Signal U506-012.

**Review by briot — COMMENTED** [review]
Patch is getting good with regards to Free I think. I will have to experiment with it a bit to test, but I think we are getting there. I realized that I had added locally a `AWS.Net.Websocket.Free(Object_Class)` which l need to remove now, it is dangerous.

I don't think this patch addresses the matter of concurrent calls to `Send` though, so we also need to work on that as a second step.

File: src/core/aws-net-websocket-registry.adb [hunk 4]
```diff
@@ -499,9 +504,11 @@ package body AWS.Net.WebSocket.Registry is
...
 
             for Id of Watched loop
-               FD_Set.Add
-                 (Result,
-                  Registered (Id).all, Registered (Id), FD_Set.Input);
+               if not Registered (Id).To_Free then
+                  FD_Set.Add
+                    (Result,
+                     Registered (Id).all, Registered (Id), FD_Set.Input);
+               end if;
```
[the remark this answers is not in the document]
TurboGit: Good point !

src/core/aws-net-websocket-registry.adb
```diff
@@ -551,6 +558,32 @@ package body AWS.Net.WebSocket.Registry is
...
+         else
+            Release_Memory (Object (WebSocket.all));
+            Unchecked_Free (WebSocket);
+         end if;
+      end Free_Or_Defer;
+
+      procedure Free_Or_Defer (WebSocket : in out Object'Class) is
+      begin
+         if Registered.Contains (WebSocket.Id) then
+            Free_Or_Defer (Registered (WebSocket.Id));
```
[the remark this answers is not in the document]
TurboGit: No, currently this must be done with Unregister. But I'll unregister WebSocket when freeing them for safety.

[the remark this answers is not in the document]
briot: Right, I think this is the safe approach. Users should never unregister on their own, simply close the socket and let AWS handle the cleanup

src/core/aws-net-websocket-registry.adb
```diff
@@ -466,7 +471,7 @@ package body AWS.Net.WebSocket.Registry is
          Timeout : Duration := Forever;
          Error   : Error_Type := Normal_Closure)
       is
-         W : Object_Class;
+         W : Object_Class := null;
```
briot, line 474: Why not, but as a pointer it is of course initialized to null already. AWS doesn't seem to have a policy of explicitly initializing pointers ?

src/core/aws-net-websocket-registry.adb
```diff
@@ -482,7 +487,9 @@ package body AWS.Net.WebSocket.Registry is
          Socket.On_Close (Message);
          Socket.Shutdown;
 
-         Unchecked_Free (W);
+         if W /= null then
+            DB.Free_Or_Defer (W);
```
briot, line 491: It would seem more usual for Free_Or_Defer to accept null (and do nothing), like unchecked_deallocation does. But this is very minor

src/core/aws-net-websocket-registry.adb
```diff
@@ -551,6 +562,37 @@ package body AWS.Net.WebSocket.Registry is
...
+      procedure Free_Or_Defer (WebSocket : in out Object_Class) is
+      begin
+         --  If WebSocket is in Sending it means that it has been
+         --  taken by the Get_Socket call. We cannot free it now, we
+         --  record this socket to be freed as soon as it is released
+         --  (Release_Socket) call.
+
+         if Sending.Contains (WebSocket.Id) then
+            WebSocket.To_Free := True;
+
```
briot, line 578: wondering whether we should reset WebSocket to null in this case as well. Users should really not try to use it anymore after calling Free_Or_Defer anyway, so this might avoid subtle bugs (where code behaves differently depending on whether we actually freed or deferred.

src/core/aws-net-websocket-registry.adb
TurboGit: Sure, that was just to make this explicit as I missed the fact that W could stay null hence the check below. But I agree this change is purely cosmetic.

src/core/aws-net-websocket-registry.adb
TurboGit: Not sure, in the other Free_Or_Defer we have:
`   Free_Or_Defer (Registered (Id));`
If you set WebSocket to null it will also nullify the pointer recorded into Registered container.

briot: But isn't that the point: we should first unregister and then free_or_defer. Otherwise, the risk is high that later someone will do additional changes after the call and not realized that it is unsafe to call the pointer which might have been freed.

TurboGit: > But isn't that the point: we should first unregister and then free_or_defer.

No the part quoted is registering the WebSocket to be freed, but we still need the handle until we actually release the memory in the object and the pointer itself. Ot there is really something that I don't understand in your proposal.

briot: I am not sure I get your point. In the case where we can actually free memory, the pointer will be reset to null, so the rest of the code cannot depend on still having the pointer afterwards. The caller to Free_Or_Defer should always expect that the pointer can be reset to null. So it seems like the code `Free_Or_Defer (Registered (Id))` has wrong expectations here.

```
      procedure Free_Or_Defer (Id : UID) is
      begin
         if Registered.Contains (Id) then
            Free_Or_Defer (Registered (Id));
         end if;
      end Free_Or_Defer;
```

This code likely should unregister the socket too. If I understand the code correctly (and that's not a given... :-) the socket will be freed as soon as it is released anyway, so we no longer need to have an entry for the socket in the registry. So perhaps something like:

```
      procedure Free_Or_Defer (Id : UID) is
         W : Object_Class;
         C : WebSocket_Set.Cursor;
      begin
         C := Registered.Find (Id);
         if Has_Element (C) then
             W := Element (C);
             Registered.Delete (C);
             Free_Or_Defer (W);
         end if;
      end Free_Or_Defer;
```

I might definitely be wrong, the code is not so easy to follow.

src/core/aws-net-websocket-registry.ads
```diff
@@ -176,10 +176,18 @@ package AWS.Net.WebSocket.Registry is
...
+   --  associated memory.
 
    function Is_Registered (Id : UID) return Boolean;
    --  Returns True if the WebSocket Id is registered and False otherwise
 
+   procedure Free (WebSocket : in out Object'Class);
+   --  Free the WebSocket when not used anymore (Either free immediatly or
+   --  register a deferred free). The difference with Close is that no message
+   --  is sent to the peer. The WebSocket is simply unregistered and memory is
+   --  released when possible.
```
briot, line 189: The documentation definitely helps here. But it raises the question why we need the two `Free` and `Close` subprograms. Why not simply have `Close` with an empty message if you do not want to send a message to the peer ? The goal is to reduce the size of the public API, perhaps to help users better understand what they are supposed to call (I know it wasn't clear to me when I started using websockets three years ago, and is still not 100% clear)

briot:
I tried running my code with the latest version of this patch, and it is crashing with a use-after-free error. To be precise, I do not use exactly the po/fix-free-ws branch because I also need the patches from #159. Here is the address-sanitizer report (I use this rather than valgrind because the latter has issues in how it handles Long_Float).
```
=================================================================
==10753==ERROR: AddressSanitizer: heap-use-after-free on address 0x6190002210b0 at pc 0x0000014c6404 bp 0x147ff1520d20 sp 0x147ff1520d18
READ of size 8 at 0x6190002210b0 thread T39
    #0 0x14c6403 in aws__net__websocket__registry__db__unregisterN aws-net-websocket-registry.adb:1117
    #1 0x14c6572 in aws__net__websocket__registry__db__unregisterP aws-net-websocket-registry.adb:1115
    #2 0x14b578f in aws__net__websocket__registry__message_reader__do_unregister aws-net-websocket-registry.adb:343
    #3 0x14b540d in aws__net__websocket__registry__message_readerTB aws-net-websocket-registry.adb:376
    #4 0x3a3b84b in system__tasking__stages__task_wrapper s-tassta.adb:1201
    #5 0x147ffcd686da in start_thread pthread_create.c:463
    #6 0x147ffb54971e in __clone

0x6190002210b0 is located 48 bytes inside of 928-byte region [0x619000221080,0x619000221420)
freed by thread T39 (message_readers) here:
    #0 0x147ffd23199f in __interceptor_free asan_malloc_linux.cpp:123
    #1 0x39caa77 in __gnat_free s-memory.adb:122
    #2 0x39e7825 in system__pool_global__deallocate s-pooglo.adb:134
    #3 0x3a1639d in system__storage_pools__subpools__deallocate_any_controlled s-stposu.adb:470
    #4 0x14bcaf6 in aws__net__websocket__registry__dbT__free_or_defer__P4246b__2.351
    #5 0x14bc76d in aws__net__websocket__registry__db__free_or_deferN__2 aws-net-websocket-registry.adb:590
    #6 0x14bcb92 in aws__net__websocket__registry__db__free_or_deferP__2 aws-net-websocket-registry.adb:574
    #7 0x14b571a in aws__net__websocket__registry__message_reader__do_free aws-net-websocket-registry.adb:325
    #8 0x14b6244 in aws__net__websocket__registry__message_reader__read_message aws-net-websocket.adb:481
    #9 0x14b5304 in aws__net__websocket__registry__message_readerTB aws-net-websocket-registry.adb:371
    #10 0x3a3b84b in system__tasking__stages__task_wrapper s-tassta.adb:1201
    #11 0x147ffcd686da in start_thread pthread_create.c:463
```

briot:
(and more tests in the testsuite, this kind of error should really be captured by the testsuite)

TurboGit:
Thanks Manu for the analysis which seems correct at first glance. I'll propose a fix for this.

TurboGit:
@briot : Hello Manu, a new version is now pushed which use UID for the message reader / sender. Hopefully this will make the code more stable.

src/core/aws-net-websocket-registry.adb
```diff
@@ -494,14 +524,18 @@ package body AWS.Net.WebSocket.Registry is
...
             for Id of Watched loop
-               FD_Set.Add
-                 (Result,
-                  Registered (Id).all, Registered (Id), FD_Set.Input);
+               declare
+                  W : constant Object_Class := Registered (Id);
+               begin
+                  if not W.To_Free then
+                     FD_Set.Add (Result, W.all, W.Id, FD_Set.Input);
+                  end if;
```
briot, line 537: When I wrote a server (not based on AWS), I was also ensuring that sockets for which we got a notification in `epoll()` were removed from the `FD_Set` until we had actually read or written everything to/from it. Otherwise, that socket keeps popping up in `epoll` which is inefficient. Is something like this necessary here (perhaps a new Issue, though)

src/core/aws-net-websocket-registry.adb
```diff
@@ -614,6 +679,21 @@ package body AWS.Net.WebSocket.Registry is
...
+      ----------------
+      -- Get_Socket --
+      ----------------
+
+      procedure Get_Socket (Id : UID; WebSocket : out Object_Class) is
+      begin
+         if Registered.Contains (Id) then
+            Sending.Insert (Id);
+
+            WebSocket := Registered (Id);
```
briot, line 691: There are two lookups in `Registered` here, better to get a `Cursor`  and check `Has_Element` on it.

briot (commented):
I will test the patch later today. I am still worried about missing exception handlers, and thus missing calls to Release_Socket.

briot:
I think we should have a `Use_Socket (UID, Callback'Access)` procedure (which I suggested earlier). It would make sure that even in case of exception, `Release_Socket` is properly called. I would also make this procedure public, so that users can take advantage of it (useful to me because I have a background task in charge of asynchronously sending messages, though I could of course use `Get_Socket` and `Release_Socket`

src/core/aws-net-websocket-registry.adb
```diff
@@ -342,16 +364,19 @@ package body AWS.Net.WebSocket.Registry is
...
 
-            Message_Queue.Get (WebSocket);
+            Message_Queue.Get (WS_Id);
+
+            DB.Get_Socket (WS_Id, WebSocket);
 
             --  A WebSocket is null when termination is requested
 
-            exit Handle_Message when WebSocket = null;
+            exit Handle_Message when WS_Id = No_UID or else WebSocket = null;
```
briot, line 379: when WS_Id is No_UID, Get_Socket will already return null, right ? So the extra test is redundant. Or else we should check for No_UID before we even call the protected procedure Get_Socket, for efficiency

briot:
With latest patch, I got the following thread-sanitizer error: 
```
WARNING: ThreadSanitizer: heap-use-after-free
  Read of size 8 at 0x7b6000060410 by thread T39:
    #0 aws__net__websocket__registry__message_readerTB aws-net-websocket-registry.adb:398
       (call to On_Close in exception handler)
    #1 system__tasking__stages__task_wrapper s-tassta.adb:1201

  Previous write of size 8 at 0x7b6000060410 by thread T39 (mutexes: write M446):
    #0 free tsan_interceptors_posix.cpp:707
    #1 __gnat_free s-memory.adb:122
    #2 system__pool_global__deallocate s-pooglo.adb:134
    #3 system__storage_pools__subpools__deallocate_any_controlled s-stposu.adb:470
    #4 aws__net__websocket__registry__dbT__free_or_defer__P4254b__2.351 <null>
    #5 aws__net__websocket__registry__db__free_or_deferN__2 aws-net-websocket-registry.adb:610
    #6 aws__net__websocket__registry__db__free_or_deferP__2 aws-net-websocket-registry.adb:594
    #7 aws__net__websocket__registry__message_reader__do_free aws-net-websocket-registry.adb:339
    #8 aws__net__websocket__registry__message_reader__read_message aws-net-websocket.adb:482
    #9 aws__net__websocket__registry__message_readerTB aws-net-websocket-registry.adb:388
    #10 system__tasking__stages__task_wrapper s-tassta.adb:1201
```

Maybe the call to Unregister should come later ?

briot:
Moving Unregister doesn't help. So maybe the exception that led us to the handler actually came from the call to Release_Socket (for sure, if this is the case we should have a different exception handler)

briot:
I added some logs to know what exception is raised, and got
```
CONSTRAINT_ERROR AWS.Net.WebSocket.Registry.WebSocket_Set.Insert: attempt to insert element already in set
0xf41518 Aws.Net.Websocket.Registry.Websocket_Set.Insert.2Xnnn at a-coorse.adb:1093
0xf6e931 Aws.Net.Websocket.Registry.Db.Get_SocketN at aws-net-websocket-registry.adb:696
0xf6ebbd Aws.Net.Websocket.Registry.Db.Get_SocketP at aws-net-websocket-registry.adb:693
0xf682e6 Aws.Net.Websocket.Registry.Message_ReaderT at aws-net-websocket-registry.adb:381
0x24e80c4 System.Tasking.Stages.Task_Wrapper at s-tassta.adb:1201
```
So indeed it is possible to `Get_Socket` to raise an exception apparently, in which case WebSocket is still null in the exception handler and we get a second exception...

src/core/aws-net-websocket-registry.adb
TurboGit: Yes, for another issue. Let's keep this for the stability of the WS support.

src/core/aws-net-websocket-registry.adb
```diff
@@ -160,10 +160,19 @@ package body AWS.Net.WebSocket.Registry is
...
+      procedure Get_Socket (Id : UID; WebSocket : out Object_Class);
+      --  Get a WebSocket Id or null if the WebSocket is not registerred
+      --  anymore.
+
+      procedure Release_Socket (WebSocket : in out Object_Class);
       --  Release a socket retrieved with Get_Socket above, this socket will be
       --  then available again.
 
+      procedure Free_Or_Defer (Id : UID);
+      procedure Free_Or_Defer (WebSocket : in out Object_Class);
```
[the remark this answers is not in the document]
TurboGit: In fact we need both as the one with object_class is called after unregister and can work with unregistered sockets.

TurboGit:
> As mentioned in previous code reviews, the simplest is to replace `Sending` with a simple flag on the socket....

Agreed, time to to that.

src/core/aws-net-websocket-registry.adb
```diff
@@ -361,13 +386,16 @@ package body AWS.Net.WebSocket.Registry is
...
 
+            DB.Release_Socket (WebSocket);
+
          exception
             when E : others =>
                Do_Unregister (WebSocket);
                WebSocket_Exception
                   (WebSocket, Exception_Message (E), Protocol_Error);
                WebSocket.On_Close (Exception_Message (E));
                WebSocket.Shutdown;
```
briot, line 397: Shutdown and On_Close are already called by WebSocket_Exception and should not be called a second time I think

TurboGit: Done here and some other places.

src/core/aws-net-websocket-registry.adb
```diff
@@ -361,13 +386,16 @@ package body AWS.Net.WebSocket.Registry is
...
+               DB.Release_Socket (WebSocket);
```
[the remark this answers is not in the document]
TurboGit: Fixed.

src/core/aws-net-websocket-registry.adb
```diff
@@ -286,11 +303,16 @@ package body AWS.Net.WebSocket.Registry is
...
+                  DB.Get_Socket (FD_Set.Get_Data (Set, K), WS);
+
+                  if WS /= null then
+                     WS.State.Errno := Error_Code (Internal_Server_Error);
+                     WS.On_Error
+                       ("WebSocket Watcher server error, "
+                        & Exception_Message (E));
+                  end if;
+
+                  DB.Release_Socket (WS);
```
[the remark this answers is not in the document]
TurboGit: Fixed.

src/core/aws-net-websocket-registry.adb
```diff
@@ -268,15 +277,23 @@ package body AWS.Net.WebSocket.Registry is
...
-                     WS := FD_Set.Get_Data (Set, K);
-                     DB.Remove (WS);
-                     Message_Queue.Add (WS);
+                     WS_Id := FD_Set.Get_Data (Set, K);
+                     DB.Get_Socket (WS_Id, WS);
+
+                     if WS /= null then
+                        DB.Remove (WS);
+                        Message_Queue.Add (WS_Id);
+                        DB.Release_Socket (WS);
```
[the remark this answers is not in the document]
TurboGit: Fixed.

TurboGit:
@briot : A new version is available with most comments resolved.

> I think we should have a `Use_Socket (UID, Callback'Access)` procedure (which I suggested earlier). It would make sure that even in case of exception, `Release_Socket` is properly called.

This is not done, cleaner indeed but still calling the Get_Socket & Release_Socket as done currently should work. I'll get on this later.

briot:
I did not include your very last changes. I was experimenting with my own, and almost managed to get my test work without errors from thread-sanitizer (but see below). The patch looks like the following (hard to send exactly, since I also have all the changes from the other PR). It also removes a number of calls to `Unregister` when this is doing along with `Free_Or_Defer` since the latter already unregisters as needed. Mostly the clean that seems to have improved things the most is when we handle the exception in message_reader, to avoid calling Shutdown a second time (when the socket might have been freed already).

```
   task body Message_Reader is

      procedure Do_Free (WebSocket : in out Object_Class);
      procedure Do_Register (WebSocket : Object_Class);
      procedure Do_Unregister (WebSocket : Object_Class);

      -------------
      -- Do_Free --
      -------------

      procedure Do_Free (WebSocket : in out Object_Class) is
      begin
         DB.Free_Or_Defer (WebSocket);
      end Do_Free;

      -----------------
      -- Do_Register --
      -----------------

      procedure Do_Register (WebSocket : Object_Class) is
      begin
         DB.Watch (WebSocket);
      end Do_Register;

      -------------------
      -- Do_Unregister --
      -------------------

      procedure Do_Unregister (WebSocket : Object_Class) is
      begin
         DB.Unregister (WebSocket);
      end Do_Unregister;

      function Read_Message is new AWS.Net.WebSocket.Read_Message
         (Receive    => DB.Receive,
          On_Success => Do_Register,
          On_Error   => Do_Unregister,
          On_Free    => Do_Free);

       WS_Id     : UID;

      -----------
      -- On_WS --
      -----------

      procedure On_WS (WebSocket : in out Object'Class) is
         WS      : Object_Class := WebSocket'Unchecked_Access;
         Message : Unbounded_String := Null_Unbounded_String;
      begin
         --  A message can be sent in multiple chunks and/or multiple
         --  frames with possibly some control frames in between text or
         --  binary ones. This loop handles those cases.

         loop
            exit when Read_Message (WS, Message);
         end loop;

      exception
         when E : others =>
            Log (Me.Error & "MANU" & E);

            --  Socket might have been freed by Read_Message
            if WS /= null then
               WebSocket_Exception (WS, Exception_Message (E), Protocol_Error);
               DB.Free_Or_Defer (WS_Id);
            end if;
      end On_WS;

   begin
      loop
         Message_Queue.Get (WS_Id);
         Use_Socket (WS_Id, On_WS'Access);
      end loop;
   end Message_Reader;

    procedure Use_Socket
       (Id       : UID;
        Callback : not null access procedure (WS : in out Object'Class))
    is
       WS : Object_Class;
    begin
       DB.Get_Socket (Id, WS);
       if WS /= null then
          begin
             Callback (WS.all);
          exception
             when others =>
                DB.Release_Socket (WS);
                raise;
          end;

          DB.Release_Socket (WS);
       end if;
    end Use_Socket;
```

The one remaining error (for now...) is that I am getting an error when a socket is finalized and freed in `Free_Or_Defer`, because `AWS.Net.Finalize` seems to be doing a potentially blocking operation (and this is called from a protected object, and we compile with the GNAT switches to raise error in this case). I haven't fully understood that one yet, and it might well be a GNAT bug (we reported a number of those already).

briot:
I have also cleaned up my own code (which was manipulating `Object_Class`) so that it now uses `Use_Socket`. That certainly helped stability too

briot:
Finally got more details on the finalization error (doing a potentially blocking call from a protected object). It shows a case of Free_Or_Defer calling itself indirectly, which should never happen.

```
system.tasking.protected_objects.entries.lock_entries_with_status at s-tpoben.adb:240
system.tasking.protected_objects.entries.lock_entries at s-tpoben.adb:212
aws.net.websocket.registry.db.free_or_defer at aws-net-websocket-registry.adb:611
aws.net.websocket.registry.free at aws-net-websocket-registry.adb:1263 (discriminator 2)
   # this Free operator will try to call Free_Or_Defer again, which is potentially blocking. This is wrong.
aws.net.websocket.free at aws-net-websocket.adb:197
aws.net.finalize at aws-net.adb:142 (discriminator 1)
   # This is the call `Free (Socket_Type'Class (Socket));`
gui.websockets.websocket_type.Finalize at a-cdlili.ads:395
   # line number is meaningless. I do not override Finalize, so this is the one from aws-net.adb
<aws__net__websocket__registry__db__free_or_deferN__2> at aws-net-websocket-registry.adb:607 (discriminator 12)
<aws__net__websocket__registry__db__free_or_deferP__2> at aws-net-websocket-registry.adb:594 (discriminator 1)
   # this is a protected procedure, should not call any potentially blocking operation
aws.net.websocket.registry.message_reader.do_free at aws-net-websocket-registry.adb:343
aws.net.websocket.registry.message_reader.read_message at aws-net-websocket.adb:482
aws.net.websocket.registry.message_reader.on_ws at aws-net-websocket-registry.adb:385
aws.net.websocket.registry.use_socket at aws-net-websocket-registry.adb:1587 (discriminator 2)
<aws__net__websocket__registry__message_readerTB> at aws-net-websocket-registry.adb:402 (discriminator 2)
system.tasking.stages.task_wrapper at s-tassta.adb:1201
```

src/core/aws-net-websocket-registry.adb
```diff
@@ -286,8 +287,16 @@ package body AWS.Net.WebSocket.Registry is
...
-                        DB.Remove (WS);
-                        Message_Queue.Add (WS_Id);
+                        --  We should never fail to ensure the WebSocket is
+                        --  properly released.
+                        begin
+                           DB.Remove (WS);
+                           Message_Queue.Add (WS_Id);
+                        exception
+                           when others =>
+                              null;
```
briot, line 297: don't we want to forward the exception to the user though ? Silently dropping exception without even logging them seems bad ?

src/core/aws-net-websocket-registry.adb
```diff
@@ -332,7 +348,10 @@ package body AWS.Net.WebSocket.Registry is
 
       procedure Do_Free (WebSocket : in out Object_Class) is
       begin
-         DB.Free_Or_Defer (WebSocket);
+         DB.Free_Or_Defer (WebSocket.Id);
+      exception
+         when others =>
+            null;
```
briot, line 354: likewise, silently dropping exception should not happen.

briot:
BTW, things are getting confused in the review tab on github. How do you guys manage that in general ? I am not that used to github, so I welcome any trick that help us keep track of things (resolved, pending, wont-fix,...)

briot:
I just realized the implementation of `Get_Socket` is wrong. I had mentioned before that it should not return the socket if it is already in use (`Sending is true`). I could not find your answer again, but I don't think this is done automatically.

But even that is not enough anyway. Instead, `Get_Socket` needs to be blocking, and wait until the socket is available (see Pat Roger's recent blog post on the standard pattern using requeue for that if you were not familiar with it already). The reason is that otherwise if we have two tasks calling `Use_Socket` concurrently (or the equivalent low-level `Get_Socket`/`Release_Socket`), then only of them will see the socket (the other will get `null` and not perform any work on it).

At it currently is implemented, it is even worse, since both tasks will receive the socket, and the first one will release it, potentially triggering a free. The second one is then accessing dangling pointers...

TurboGit:
> I just realized the implementation of `Get_Socket` is wrong. I had mentioned before that it should not return the socket if it is already in use (`Sending is true`). I could not find your answer again, but I don't think this is done automatically.

Well the "fix" to use a boolean in WebSocket object is wrong because the state (boolean) is not shared across all copies of WebSocket object. Either we should use an access to a boolean or revert to using a container.

TurboGit:
Or move To_Free and Sending boolean into the Internal_State record which was made for this !

briot:
Not sure I followed that. The socket uses the same UID everywhere right, so the container for Sending (which maps from this UID to the boolean) would have the same issue ? That's a part of the code I did not look into, so I'll certainly trust you there.

Still, that would not solve the need to make `Get_Socket` blocking

briot:
Also, what copies of the socket ? Internally, you use Object_Class, so this is a pointer to a single `Object`. This object must be the same all the type, since it might be a user-defined type with its own data (that's more related to the second PR, but we should not break that in this PR).

TurboGit:
@briot : With the latest change I feel that we are really in better position now. My feeling is that we could merge this now and see if some other fixes are to be applied. Is that ok with you ?

briot:
I have to review your change (missed it when you committed it). As I remember there were a number of points of discussion still pending, but things have getting a bit confusing in this PR indeed. I'll review and test the patch tomorrow, will let you know

src/core/aws-net-websocket-registry.adb
```diff
@@ -225,10 +235,9 @@ package body AWS.Net.WebSocket.Registry is
       Signal      : Boolean := False;    -- Transient signal, release Not_Emtpy
       S_Signal    : Boolean := False;    -- Shutdown is in progress
       New_Pending : Boolean := False;    -- New pending socket
+      New_State   : Boolean := False;    -- A sokcet has been released
```
briot, line 238: sokcet => socket

briot (commented):
I am not really comfortable with all the "exception when others => null" which simply ignore exceptions. I wonder what these will hide...

There are a number of cases where we do `DB.Unregister(...); DB.Free_Or_Defer(...)` I thought the latter already handle de-registration, so it seems the former call is unneeded (and inefficient since we need to lock the PO).

I still would like to argue for adding a `Use_Socket` function that will automatically take care of the Get_Socket+Release_Socket. It will cleanup the AWS code a bit, but more importantly it allows users to do things safely in their own code. In my case, I have a queue of messages I want to send to users, but this is done asynchronously. So the queue has the socket UID and the message to send. The only way to be task-safe is to use get_socket+release_socket, but those are currently hidden to users. A Use_Socket that encapsulate that is also safer to use...
The code looks like
```
   procedure Use_Socket
      (Id       : UID;
       Callback : not null access procedure (WS : in out Object'Class))
   is
      WS : Object_Class;
   begin
      DB.Get_Socket (Id, WS);
      if WS /= null then
         begin
            Callback (WS.all);
         exception
            when others =>
               DB.Release_Socket (WS);
               raise;
         end;

         DB.Release_Socket (WS);
      end if;
   end Use_Socket;
```

Thanks for the clean up of `Sending`.

I am testing the patch (merging takes me a very long time since I have a number of additional patches, mostly from #159, so I have a local copy of the AWS files to which I backport the changes. That doesn't look like the most efficient way to do things though. Once we merge this one and perhaps #159, I'll try to get back to an easier setup.

briot:
My testsuite never completed, full of 
```
14:59:19.561  [GUI.WEBSOCKETS@ERROR/00001458AA649000] On_Close PROGRAM_ERROR System.Tasking.Protected_Objects.Operations.Protected_Entry_Call: potentially blocking operation
0x1d7f09e System.Tasking.Protected_Objects.Operations.Protected_Entry_Call at s-tpobop.adb:512
0xe6cf0a Aws.Net.Websocket.Registry.Use_Socket at aws-net-websocket-registry.adb:1663
0x1b1422a Gui.Websockets.Active_Websockets.RemoveN at gui-websockets.adb:597
0x1b14465 Gui.Websockets.Do_Close at gui-websockets.adb:586
0xe66420 Aws.Net.Websocket.Registry.Message_Reader.On_Ws at aws-net-websocket.adb:482
0xe6cf2c Aws.Net.Websocket.Registry.Use_Socket at aws-net-websocket-registry.adb:1666
0xe6d087 Aws.Net.Websocket.Registry.Message_ReaderT at aws-net-websocket-registry.adb:421
0x1d7bdad System.Tasking.Stages.Task_Wrapper at s-tassta.adb:1201
[/lib/x86_64-linux-gnu/libpthread.so.0]
0x1458b1dec6d9
[/lib/x86_64-linux-gnu/libc.so.6]
```

This is the call to the entry `Get_Socket` (used to be a protected procedure). As I mention before, I think making this an entry was correct though, so the logic error is in my code. That however will delay the testing... Will come back to you shortly. Worth noting that this change might impact users though (again, I think the new behavior is the correct one)

briot:
The implementation of `Get_Socket` is wrong I think (blocking forever).
It has a barrier on the `New_State` entry, which you immediately reset to `False`. So imagine the following scenario:

* task 1 gets the socket 1
* task 2 also gets the socket 1, and blocks (it is `Sending`)
* task 3 gets the socket 2
* task 4 also gets the socket and blocks (it is `Sending`)
* then task 3 releases the socket 2. This sets New_State to True, which unblocks either task 2 or task 4. Let's assume task 2. It starts executing Get_Socket, sets New_State to False, then discovers socket 1 is still Sending, so goes back to sleep. But task 4 cannot enter Get_Socket, since New_State is now False !

The proper implementation is to replace New_State with `Available : Natural`. In Release_Socket, you set Available to `Get_Socket'Count` so that all currently blocked tasks will get a chance to unblock. Again, see Pat's latest blog https://blog.adacore.com/on-the-benefits-of-families

briot:
The implementation of `Get_Socket` is wrong I think (blocking forever).
It has a barrier on the `New_State` entry, which you immediately reset to `False`. So imagine the following scenario:

* task 1 gets the socket 1
* task 2 also gets the socket 1, and blocks (it is `Sending`)
* task 3 gets the socket 2
* task 4 also gets the socket and blocks (it is `Sending`)
* then task 3 releases the socket 2. This sets New_State to True, which unblocks either task 2 or task 4. Let's assume task 2. It starts executing Get_Socket, sets New_State to False, then discovers socket 1 is still Sending, so goes back to sleep. But task 4 cannot enter Get_Socket, since New_State is now False !

The proper implementation is to replace New_State with `Available : Natural`. In Release_Socket, you set Available to `Get_Socket'Count` so that all currently blocked tasks will get a chance to unblock. Again, see Pat's latest blog https://blog.adacore.com/on-the-benefits-of-families

briot:
And this really shows the AWS testsuite for websocket is lacking... I believe the code I sent in https://github.com/AdaCore/aws/pull/209#issuecomment-832584313 shows this issue

briot:
Hi Pascal, any news on this patch ? did you manage to fix the issue with requeue ?

TurboGit:
Hi Manu, no progress so far but work is restarting on this issue.

TurboGit:
Hi Manu, with your reproducer we cannot reproduce issues. I'm wondering if you are using some specific options (we use only -fsanitize=thread) ? Or maybe we are not looking at the right place or for the right report ? Can you describe more your procedure ? TIA.

briot:
We use the same simple switch that Dmitriy proposed in #214, namely `-fsanitize=thread`.
One big difference is our use of the configuration pragma `Detect_Blocking` which raises errors on potentially blocking operations done from protected objects (not sure it plays a role here).

The latest error I reported in `Get_Socket` might not have been detected by the test. I discovered it when I was looking at why things are executing much more slowly now, and then it was by reading the code. The scenario I described then could be used to write another dedicated test. If you run it long enough with enough tasks, but it isn't clear how to detect from an automatic test that things are blocking for too long.

I am assuming you tried the test I sent (https://github.com/AdaCore/aws/pull/209#issuecomment-832584313) without any of the patches in this discussion (which did improve things, no mistake, even if not perfect yet).

I'll try double-check again

briot:
Hi Pascal and Dmitry

Any news on this issue ? We are almost there, the patch is almost done and definitely improves things. I think mostly the implementation of `Get_Socket` is wrong, based on my message from May 27.

anisimkov:
I did not see the prove that current AWS sources is not thread safe. Manu, if I miss something, could you show reproducer ?

TurboGit:
> We are almost there, the patch is almost done and definitely improves things.

Agreed, so may I propose to merge this part and see where we are and get to last issue if any. At this point the discussion is long, the path is not trivial, so I think moving forward by step will certainly help converging. How does that sound to you Manu?

briot:
Agreed the discussion is indeed too long, Github is not quite up to handling such long patches it seems. I'll open a separate issue afterwards.
Give me an hour or so, I am trying to come up with a reproducer for the initial issue at least

anisimkov:
I don't think we can merge this pull request. I see a lot of

briot:
Here is a reproducer that shows issues in thread sanitizer. To be honest, I am not sure whether those are related to this issue (#209) or to the one reported by Maxim in #138, though both need to be fixed anyway.

Dmitry, just take the time to run the demos/websockets under any sanitizer/valgrind tool you want, and you'll see the issue reported by Maxim in #138. This might need to be fixed before the issues in #209 are apparent (I have run locally with workarounds for #138 for a few years now, so indeed those are no longer an issue for me)

As the long discussion here shows, there are a lot of concurrency issues in the current websocket code that we can see just by reading the code. Pascal's patch fixes a lot of them (except the one in Get_Socket I reported in May -- I kind of lost the context from then, but we'll get to that after Pascal merges the initial set of cleanups)

briot:
```
with AWS.Server;
package Client is
   task type Client_Type (Id : Positive) is
      entry Terminated;
   end Client_Type;

   type Client_Access is access Client_Type;
   type Client_Array is array (Natural range <>) of Client_Access;

   Finish : Boolean := False with Atomic;
   --  Request termination of tasks

   Server : AWS.Server.HTTP;

end Client;
with Ada.Exceptions;
with Ada.Numerics.Float_Random;
with Ada.Text_IO;
with AWS.Net.WebSocket;
with AWS.Server.Status;
package body Client is

   task body Client_Type is
      use AWS.Net.WebSocket;
      Sock : AWS.Net.WebSocket.Object_Class;
      Gen  : Ada.Numerics.Float_Random.Generator;
   begin
      Ada.Numerics.Float_Random.Reset (Gen);
      while not Finish loop
         Sock := new AWS.Net.WebSocket.Object;  --  client-side socket
         AWS.Net.WebSocket.Connect
            (Sock.all,
             AWS.Server.Status.Local_URL (Server) & "/streams");

         Sock.Send ("some message" & Id'Image);

         Sock.Close ("no longer needed" & Id'Image);
         --  AWS.Net.WebSocket.Free (Sock);

         delay 3.0 * Duration (Ada.Numerics.Float_Random.Random (Gen));
      end loop;

      accept Terminated;

   exception
      when E : others =>
         Ada.Text_IO.Put_Line (Ada.Exceptions.Exception_Information (E));
         accept Terminated;
   end Client_Type;
end Client;
with Ada.Text_IO;
with Ada.Unchecked_Deallocation;
with AWS.Config.Set;
with AWS.Net.WebSocket.Registry.Control;
with AWS.Server;
with AWS.Services.Dispatchers.URI;
with AWS.Status;
with Client;

procedure Main is

   procedure Unchecked_Free is new Ada.Unchecked_Deallocation
      (Client.Client_Type, Client.Client_Access);

   type My_Socket is new AWS.Net.WebSocket.Object with null record;
   overriding procedure On_Message (Self : in out My_Socket; Msg : String);

   overriding procedure On_Message (Self : in out My_Socket; Msg : String) is
   begin
      Ada.Text_IO.Put_Line (Msg);
   end On_Message;

   function Create_Socket
     (Socket : AWS.Net.Socket_Access;
      Request : AWS.Status.Data) return AWS.Net.WebSocket.Object'Class
   is
      --  strange code here to create our own socket, see issue 138
   begin
      return My_Socket'   --  server-side socket
         (AWS.Net.WebSocket.Object (AWS.Net.WebSocket.Create (Socket, Request))
          with null record);
   end Create_Socket;

   Arr     : Client.Client_Array (1 .. 60);
   Conf    : AWS.Config.Object := AWS.Config.Get_Current;
   Handler : AWS.Services.Dispatchers.URI.Handler;
begin
   AWS.Config.Set.Reuse_Address (Conf, True);
   AWS.Net.WebSocket.Registry.Register
      ("/streams", Create_Socket'Unrestricted_Access);
   AWS.Server.Start
      (Web_Server => Client.Server,
       Config     => Conf,
       Dispatcher => Handler);
   AWS.Net.WebSocket.Registry.Control.Start;

   for A in Arr'Range loop
      Arr (A) := new Client.Client_Type (Id => A);
   end loop;

   delay 20.0;

   Client.Finish := True;

   for A in Arr'Range loop
      Arr (A).Terminated;

      while not Arr (A)'Terminated loop
         delay 0.0001;
      end loop;

      Unchecked_Free (Arr (A));
   end loop;

   AWS.Net.WebSocket.Registry.Control.Shutdown;
   AWS.Server.Shutdown (Client.Server);
end Main;
with "aws.gpr";
project Default is
   for Main use ("main.adb");

   package Compiler is
      for Switches ("Ada") use
         ("-g", "-fsanitize=thread", "-fno-omit-frame-pointer");
   end Compiler;

   package Linker is
      for Switches ("Ada") use ("-fsanitize=thread"); --  , "-fuse-ld=gold");
   end Linker;
end Default;
make setup LAL=false XMLADA=false
make
make install prefix=`pwd`/install
cd reproducer
GPRPATH=`pwd`/../install/lib/gpr gprbuild -Pdefault.gpr
```

briot:
(I have deleted my example, I think it contained errors -- still working on a reproducer)
The demos/websockets shows at least one error in thread sanitizer, as a starting point

TurboGit:
> I don't think we can merge this pull request. I see a lot of
> +Execution of ./turl2 terminated by unhandled exception
> +raised PROGRAM_ERROR : finalize/adjust raised exception

So we have to do some clean-up first indeed. But still I'd like to move forward on this. Many improvements have been made in this PR.

briot:
An actual reproducer now. It shows some sanitizer errors in close_socket, which I think are related to this thread. It also shows PROTOCOL_ERROR, which I think are likely errors in my test, though I must admit it isn't obvious to me right now.

```
with Ada.Unchecked_Deallocation;
with AWS.Config.Set;
with AWS.Net.WebSocket.Registry.Control;
with AWS.Server;
with AWS.Services.Dispatchers.URI;
with AWS.Status;
with Client;

procedure Main is
   procedure Unchecked_Free is new Ada.Unchecked_Deallocation
      (Client.Client_Type, Client.Client_Access);

   Arr     : Client.Client_Array (1 .. 160);
   Conf    : AWS.Config.Object := AWS.Config.Get_Current;
   Handler : AWS.Services.Dispatchers.URI.Handler;
begin
   AWS.Config.Set.Reuse_Address (Conf, True);

   AWS.Net.WebSocket.Registry.Control.Start;
   AWS.Net.WebSocket.Registry.Register
      ("/streams", Client.Create_Socket'Access);
   AWS.Server.Start
      (Web_Server => Client.Server,
       Config     => Conf,
       Dispatcher => Handler);

   for A in Arr'Range loop
      Arr (A) := new Client.Client_Type (Id => A);
   end loop;

   delay 20.0;

   Client.Finish := True;

   for A in Arr'Range loop
      Arr (A).Terminated;

      while not Arr (A)'Terminated loop
         delay 0.0001;
      end loop;

      Unchecked_Free (Arr (A));
   end loop;

   AWS.Net.WebSocket.Registry.Control.Shutdown;
   AWS.Server.Shutdown (Client.Server);
end Main;
with AWS.Net.WebSocket;
with AWS.Server;
with AWS.Status;
package Client is
   task type Client_Type (Id : Positive) is
      entry Terminated;
   end Client_Type;

   type Client_Access is access Client_Type;
   type Client_Array is array (Natural range <>) of Client_Access;

   Finish : Boolean := False with Atomic;
   --  Request termination of tasks

   Server : AWS.Server.HTTP;

   --  Server-side sockets
   type My_Socket is new AWS.Net.WebSocket.Object with null record;
   overriding procedure On_Message (Self : in out My_Socket; Msg : String);
   overriding procedure On_Close (Self : in out My_Socket; Msg : String);
   overriding procedure On_Error (Self : in out My_Socket; Msg : String);

   function Create_Socket
     (Socket : AWS.Net.Socket_Access;
      Request : AWS.Status.Data) return AWS.Net.WebSocket.Object'Class;

end Client;
with Ada.Unchecked_Deallocation;
with Ada.Exceptions;
with Ada.Numerics.Float_Random;
with Ada.Text_IO;
with AWS.Net.WebSocket.Registry;
with AWS.Server.Status;
package body Client is
   Debug : constant Boolean := False;

   type Client_Socket is new AWS.Net.WebSocket.Object with null record;
   overriding procedure On_Message (Self : in out Client_Socket; Msg : String);

   overriding procedure On_Message
      (Self : in out Client_Socket; Msg : String) is
   begin
      Ada.Text_IO.Put_Line ("Client received" & Msg);
   end On_Message;

   procedure Unchecked_Free is new Ada.Unchecked_Deallocation
      (AWS.Net.WebSocket.Object'Class, AWS.Net.WebSocket.Object_Class);

   task body Client_Type is
      use AWS.Net.WebSocket;
      Sock : AWS.Net.WebSocket.Object_Class;
      Gen  : Ada.Numerics.Float_Random.Generator;
   begin
      Ada.Numerics.Float_Random.Reset (Gen);
      while not Finish loop
         Sock := new Client_Socket;  --  client-side socket
         AWS.Net.WebSocket.Connect
            (Sock.all,
             AWS.Server.Status.Local_URL (Server) & "/streams");

         Sock.Send ("some message" & Id'Image);

         delay 0.1 * Duration (Ada.Numerics.Float_Random.Random (Gen));

         Sock.Close ("no longer needed" & Id'Image);
         Unchecked_Free (Sock);

      end loop;

      accept Terminated;

   exception
      when E : others =>
         Ada.Text_IO.Put_Line (Ada.Exceptions.Exception_Information (E));
         accept Terminated;
   end Client_Type;

   overriding procedure On_Message (Self : in out My_Socket; Msg : String) is
   begin
      if Debug then
         Ada.Text_IO.Put_Line ("On_Message: " & Msg);
      end if;

      --  Let all recipients know about the new message.
      --  How efficient is that ? :-)
      AWS.Net.WebSocket.Registry.Send
         (AWS.Net.WebSocket.Registry.Recipient'
            (AWS.Net.WebSocket.Registry.Create ("/streams")),
          "someone joined the party and sent a message " & Msg);
   end On_Message;

   overriding procedure On_Error (Self : in out My_Socket; Msg : String) is
   begin
      if Debug then
         Ada.Text_IO.Put_Line ("On_Error: " & Msg);
      end if;
   end On_Error;

   overriding procedure On_Close (Self : in out My_Socket; Msg : String) is
   begin
      if Debug then
         Ada.Text_IO.Put_Line ("On_Close: " & Msg);
      end if;
   end On_Close;

   function Create_Socket
     (Socket : AWS.Net.Socket_Access;
      Request : AWS.Status.Data) return AWS.Net.WebSocket.Object'Class
   is
      --  strange code here to create our own socket, see issue 138
   begin
      if Debug then
         Ada.Text_IO.Put_Line ("Create_Socket");
      end if;
      return My_Socket'   --  server-side socket
         (AWS.Net.WebSocket.Object (AWS.Net.WebSocket.Create (Socket, Request))
          with null record);
   end Create_Socket;

end Client;
``

The project file is
```
with "aws.gpr";
project Default is
   for Main use ("main.adb");

   package Compiler is
      for Switches ("Ada") use
         ("-O2", "-fsanitize=thread", "-fno-omit-frame-pointer");
   end Compiler;

   package Linker is
      for Switches ("Ada") use ("-fsanitize=thread"); --  , "-fuse-ld=gold");
   end Linker;
end Default;
```
And here is how I compiled a fresh clone of AWS:
```
make setup LAL=false XMLADA=false THREAD_SANITIZER=true
make
make install prefix=`pwd`/install
cd reproducer
GPRPATH=`pwd`/../install/lib/gpr gprbuild -Pdefault.gpr
```

anisimkov:
Manu, your example gives strange output at the end even without thread sanitizer

raised AWS.CLIENT.PROTOCOL_ERROR : �;someone joined the party and sent a message some message 35HTTP/1.1 101 Switching Protocols

raised AWS.CLIENT.PROTOCOL_ERROR : �;someone joined the party and sent a message some message 35HTTP/1.1 101 Switching Protocols

raised AWS.CLIENT.PROTOCOL_ERROR : Invalid accept from server

raised AWS.CLIENT.PROTOCOL_ERROR : �;someone joined the party and sent a message some message 37HTTP/1.1 101 Switching Protocols

double free or corruption (out)
Aborted (core dumped)

briot:
Maybe you have an idea what's wrong. I did not find anything obvious in my code, so this might be early freed memory in AWS for instance, which is kind of what we are looking for here

anisimkov:
Ok. I'll try to understand.

PR #217, opened 2021-06-30
Makefile: split tools/library build targets for cross and native

pmderodat:


TurboGit (approved):
Looks good ! So if one want to build the tools & lib with different version of the compiler it is required to use build-tool-native & build-lib-native separately with different GPRBUILD set. Simplest that what I would have thought :)

pmderodat:
Actually, what we will do in production scripts is to change `PATH` between the two `make` calls, but that’s the same idea. :-) Thank you for the review!

PR #218, opened 2021-07-08
Remove Encode parameter from URL package

anisimkov:
It is not used in AWS internally and can be easy replaced by Encode call
from the same package.


TurboGit (approved):
Sounds good, thanks !

PR #219, opened 2021-07-12
Common output for RedHat ES v7 32bit

anisimkov:
U711-002

0245_errcon test output became common for the RedHat ES version 7 32bit.


anisimkov:
I don't know why rhES8 kept like before, but rhES7 changed. I do not think we need to investigate it.

TurboGit (approved):
Good to me, thanks !

PR #221, opened 2021-07-13
Fix stack overflow

anisimkov:
Use max frame size on send file.

Part of S507-051.


TurboGit (approved):
Indeed, a bad thinko, nice catch. Thanks!

PR #223, opened 2021-07-15
Separate hpack encoder and decoder tables

anisimkov:
in accordance with the RFC 7541 section 2.2.

Part of S507-051.


TurboGit (approved):
Good catch ! Thanks.

PR #226, opened 2021-07-17
Upload delay on low Flow_Control_Window

anisimkov:
Fixes failure on Chromium when uploading 10M size image.

* src/core/aws-server-protocol_handler_v2.adb
(Handle_Control_Frame): Do not update all streams on the receive connection
 level WINDOWS_UPDATE frame.
Exit from Answers processing on low Flow_Control_Window to be able to get
WINDOWS_UPDATE frames.

Part of S507-051.


TurboGit (approved):
Looks good to me, thanks a lot Dmitriy for this hard work.

PR #227, opened 2021-07-19
Makefile: split tools/library install targets for cross and native

pmderodat:


brobecke (commented):
I am wondering if adding "install-clean" to the dependencies of all 4 new install-*-* rules is what we want here. I think it might be causing problems uninstalling stuff you don't want uninstalled, if you're not careful.

Shouldn't we instead keep this as the first dependency of install-native and install-cross?

What do you think, @TurboGit ?

TurboGit:
Agreed, to me `install-clean` should the first dependency on `install-cross` and `install-native` targets.

TurboGit (changes requested):
Fix install-clean. TIA.

pmderodat:
Done, thank you for the suggestion!

Makefile
```diff
@@ -258,29 +258,37 @@ endif
...
 		$(GPR_OTHER) --build-name=$(OTHER_LIBRARY_TYPE) aws.gpr
 endif
 
-install-cross: install-clean
+install-tools-native:
+	$(GPRINSTALL) $(GPROPTS) $(GPRINST_OPTS) $(GPR_STATIC) --mode=usage \
+		--build-name=$(DEFAULT_LIBRARY_TYPE) \
+		--install-name=aws tools/tools.gpr
+
+install-native: install-lib-native install-tools-native
```
brobecke, line 274: Did you mean to add an `install-clean` here?

brobecke (changes requested):

Makefile
pmderodat: Ah yes, sorry… Should be fixed now, thanks!

brobecke (approved):
Looks good to me, Pierre-Marie. Thank you.

brobecke:
@TurboGit IIUC the GitHub log correctly, the change is still blocked pending your approval (which is not necessarily a bad thing in this case). Would you mind reviewing too?

Thank you!

TurboGit (approved):
Looks good now, thanks !

pmderodat:
Great, thank you both for the review!

PR #228, opened 2021-07-19
Complete HTTP2 log

anisimkov:
Part of S507-051.


src/http2/aws-http2-stream.ads
```diff
@@ -133,6 +142,9 @@ private
    function Flow_Control_Window (Self : Object) return Integer is
      (Self.Flow_Control_Window);
 
+   function Length_Sent (Self : Object) return Stream_Element_Count is
```
TurboGit, line 145: I would rename this to Bytes_Sent and the field in the record too.

PR #231, opened 2021-07-27
Fix HPACK encoder

anisimkov:
HPACK Encoder was not able to use dynamic tables.
Optimize HPACK Decoder.
No need Rank and Index together. For the Static table we need Index.
For the Dynamix table we need Rank only.

Part of S507-051.


TurboGit (approved):
Thanks ! I'll reformat the main.adb.

PR #233, opened 2021-07-27
Ensure that the AWS version is always ending with vx.x.

TurboGit:
This makes this test more stable across all version naming.

For RC11-007 & U726-036.

### DIFF
--- a/regtests/0192_client_headers/client_headers.adb
+++ b/regtests/0192_client_headers/client_headers.adb
@@ -1,7 +1,7 @@
 ------------------------------------------------------------------------------
 --                              Ada Web Server                              --
 --                                                                          --
---                     Copyright (C) 2008-2017, AdaCore                     --
+--                     Copyright (C) 2008-2018, AdaCore                     --
 --                                                                          --
 --  This is free software;  you can redistribute it  and/or modify it       --
 --  under terms of the  GNU General Public License as published  by the     --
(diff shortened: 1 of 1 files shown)


regtests/0192_client_headers/client_headers.adb
```diff
@@ -1,7 +1,7 @@
 ------------------------------------------------------------------------------
 --                              Ada Web Server                              --
 --                                                                          --
---                     Copyright (C) 2008-2017, AdaCore                     --
+--                     Copyright (C) 2008-2018, AdaCore                     --
```
anisimkov, line 4: 2018 ?

anisimkov (changes requested):

TurboGit:
It's a cherry pick (as-is) of another commit from 2018.

regtests/0192_client_headers/client_headers.adb
TurboGit: It's a cherry pick (as-is) of another commit from 2018.

anisimkov (approved):
I think we should cherry-pick it into all branches, at least into master to do not repeat it on each new LTS release.

anisimkov:
> I think we should cherry-pick it into all branches, at least into master to do not repeat it on each new LTS release.

I see the master already have it.

TurboGit:
Exactly : eeb5dedd3fd

PR #240, opened 2021-08-02
In header

anisimkov:


demos/upload/upload_cb.adb
```diff
@@ -35,16 +35,54 @@ package body Upload_CB is
...
+      procedure Message_Body_To_File (Filename : String);
+      --  Write message body to file
+
+      --------------------------
+      -- Message_Body_To_File --
+      --------------------------
+
+      procedure Message_Body_To_File (Filename : String) is
+         use Ada.Streams;
+         File : Stream_IO.File_Type;
```
TurboGit, line 48: align on :

src/http2/aws-http2-frame-data.adb
```diff
@@ -130,11 +162,18 @@ package body AWS.HTTP2.Frame.Data is
...
+        or else
+          (Self.Has_Flag (Padded_Flag)
+           and then Length_Type (Self.Data.D.Pad_Length)
+                      + Self.Data.D.Pad_Length'Size / 8
+                    >= Self.Header.H.Length)
+      then
          return C_Protocol_Error;
       else
-         return C_No_Error;
+         return Frame.Object (Self).Validate (Settings);
```
TurboGit, line 176: Looks strange to validate the parent frame object here. Should should have already been done before calling this Validate or if not it should be added in another place. But the Validate of the Data frame should only be looking at the actual data frame.

TurboGit (changes requested):

src/http2/aws-http2-frame-data.adb
anisimkov: it is the actual frame validation, but another properties of the frame.

demos/upload/upload_cb.adb
anisimkov: Done

src/http2/aws-http2-frame-data.adb
TurboGit: Yes, but you must not call this from here. The Validate on a specific frame must be be calling parent validation to validate headers. This is not done in all other frames kind. If this is needed maybe do that for all frames kind in Is_Valid ?

anisimkov: <<The Validate on a specific frame must be be calling parent validation to validate headers.>> It is the Specific frame we are calling for. <<This is not done in all other frames kind.>> It we did not do it for all other kind of frames, it was by mistake.
<<If this is needed maybe do that for all frames kind in Is_Valid ?>> Is_Valid is just wrapper/helper around the Validate.

TurboGit (approved):
Ok, let's go with that. Thanks.

PR #242, opened 2021-08-04
Another method to approve client upload

anisimkov:
Better fit for HTTP/2 implementation.

Part of S507-051.

### DIFF
--- a/src/core/aws-response.ads
+++ b/src/core/aws-response.ads
@@ -272,6 +272,26 @@ package AWS.Response is
    --  with the request. This is used by the callback's chain in the
    --  dispatchers and should not be used by users.
 
+   function Continue return Data with
+     Post => Status_Code (Continue'Result) = Messages.S100
+             and then Mode (Continue'Result) = No_Data;
+   --  Returns an empty message (Data_Mode = No_Data and Status_Code is 100).
+   --  It is to control the client data upload.
+   --
+   --  If upload data size is known from Content-Length header and less than
+   --  Upload_Size_Limit configuration parameter then the client message body
+   --  arrived at once to the dispatcher handler. User can check this by
+   --  calling AWS.Status.Is_Body_Uploaded.
+   --
+   --  If upload data size is unknown or more than Upload_Size_Limit then
+   --  Is_Body_Uploaded returns False and user is able to allow or disable the
+   --  client data upload.
+   --
+   --  If user returns Continue response from dispatcher handler, then next
+   --  time the dispatcher handler will be called with uploaded body from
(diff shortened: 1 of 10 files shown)


src/core/aws-response.ads
```diff
@@ -272,6 +272,22 @@ package AWS.Response is
...
+   --  It is to control the client data upload. If upload data size is known
+   --  from Content-Length header and less than Upload_Size_Limit configuration
+   --  parameter than the client message body arrived at once to the dispatcher
+   --  handler. User can understand it with Is_Body_Uploaded call from package
+   --  AWS.Status. If upload data size is unknown or more than
+   --  Upload_Size_Limit than the Is_Body_Uploaded returns False and user is
+   --  able to allow or disable the client data upload. If user returns
+   --  Continue response from dispatcher handler, then next time the dispatcher
+   --  handler will be called with uploaded body from client. If user returns
+   --  some another responce than client body upload will be terminated and
```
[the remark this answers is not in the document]
anisimkov: 2 "than" corrected to "then"

src/core/aws-response.ads
```diff
@@ -272,6 +272,22 @@ package AWS.Response is
    --  with the request. This is used by the callback's chain in the
    --  dispatchers and should not be used by users.
 
+   function Continue return Data with
+     Post => Status_Code (Continue'Result) = Messages.S100
+             and then Mode (Continue'Result) = No_Data;
+   --  Returns an empty message (Data_Mode = No_Data and Status_Code is 100).
```
TurboGit, line 278: I'm proposing this text:

```
--  Returns an empty message (Data_Mode = No_Data and Status_Code is 100).
--  It is to control the client data upload.
--
--  If upload data size is known from Content-Length header and less
--  than Upload_Size_Limit configuration parameter then the client
--  message body arrived at once to the dispatcher handler. User can
--  check this by calling AWS.Status.Is_Body_Uploaded.
--
--  If upload data size is unknown or more than Upload_Size_Limit then
--  Is_Body_Uploaded returns False and user is able to allow or
--  disable the client data upload.

--  If user returns Continue response from dispatcher handler, then
--  next time the dispatcher handler will be called with uploaded body
--  from client. If user returns some other responses then client body
--  upload will be terminated and ignored.
```

TurboGit (commented):
A proposal for the comment. Apart from that it looks good.

src/core/aws-response.ads
anisimkov: Done

TurboGit (approved):
Thanks

PR #245, opened 2021-08-11
Ensure input buffer for ZLib stream is bigger than output one.

TurboGit:
This force a one byte ahead read of the stream and properly detect
the end of compressed stream which can possibly contains padding
at the end.

For U727-039.


anisimkov:
I think we should work right in any size of buffers. I guess it is gzip footer does not allow to detect end of stream. I'll try to understand.

TurboGit:
Better & proper fix put in place.

PR #247, opened 2021-08-18
Fix context of request and responce

anisimkov:
The right context of the HTTP/2 request and responce is not in the TCP
connection but in the HTTP/2 stream.
To check this put 2 few megabites length files into web_elements directory,
run demos/wps executable, and then call

h2load https://127.0.0.1:4433/{big-file-1,big-file-2} -v -w 14 -m 2 -n 8

TN: S507-051


TurboGit (approved):
A nice one, congrats !

PR #252, opened 2021-09-15
removed GNATcheck warnings

pjljvandelaar:
removed GNATcheck warnings related to rule 11.1.5.6. Binary_Case_statements

Note 1. The changed statements have been pretty printed which might have caused a few more changes than strictly necessary.
Note 2. All subrules (like replace `not (x in y)` by `x not in y`) to get maximally readable code are applied on files with changed case statements. These subrules sometime also change other parts of the files not related to the original / changed case statements.

Note 3. I am curious what the build pipeline produces on other platforms than I have available.

### DIFF
--- a/config/setup/xoscons.adb
+++ b/config/setup/xoscons.adb
@@ -33,6 +33,7 @@
 
 --  The generated files are UNIT_NAME.ads and UNIT_NAME.h
 
+with Ada.Assertions;
 with Ada.Characters.Handling;    use Ada.Characters.Handling;
 with Ada.Command_Line;           use Ada.Command_Line;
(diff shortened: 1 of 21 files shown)


config/setup/xoscons.adb
```diff
@@ -353,61 +351,47 @@ procedure XOSCons is
          Info.Line_Number :=
            Integer (Parse_Int (Line (Index1 .. Index2 - 1), CNU).Abs_Value);
 
-         case Info.Kind is
```
pjljvandelaar, line 356: @TurboGit Although I understand your doubts with the case to if statement rule. This is an example I find definitely more readable.

PR #255, opened 2021-09-24
po/client http2

TurboGit:
Not to be merged yet. Review welcomed.


TurboGit:
@anisimkov : I have removed the "draft" status on this PR which should be ready for review. TIA.

anisimkov (changes requested):
looks like typo in commit message "All are supported properly expept when there is attachments."

anisimkov (changes requested):
<<Add routine Get_Content to retrieve the steamed headers. >> maybe stReamed ?

anisimkov (changes requested):
I think no need to use Utils.Normalize_Lower in  src/http2/aws-http2-message.adb because it is only HTTP2 package, better to use just To_Lower here,

TurboGit:
@anisimkov : new version with correction pushed.

PR #256, opened 2021-09-27
Use shared pointers to avoid use memory after deallocation

anisimkov:


TurboGit (approved):
Looks good to me. Thanks!

PR #260, opened 2021-10-06
Easy to switch default HTTP protocol

anisimkov:
TN: S507-051

### DIFF
--- a/src/core/aws-client.ads
+++ b/src/core/aws-client.ads
@@ -67,6 +67,8 @@ package AWS.Client is
    --  Number of time a data is requested from the Server if the first
    --  time fails.
 
+   HTTP_Default : HTTP_Protocol renames HTTPv1;
+
    --------------
    -- Timeouts --
    --------------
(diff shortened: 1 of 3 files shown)


TurboGit:
Not sure about this. It means that every calls using HTTP/1 today will automatically be changed to HTTP/2 when we will be changing the default. For server part it makes sense as it supports both protocols. But for the client side we either do an HTTP/1 or and HTTP/2 call. If the server is not ready (supporting) HTTP/2 the same program will fail !

anisimkov:
OK. I made separate HTTP_Default for client part, not depend on server part.

src/core/aws-client.ads
```diff
@@ -271,24 +273,25 @@ package AWS.Client is
...
+      Pwd          : String          := No_Data;
+      Proxy        : String          := No_Data;
+      Proxy_User   : String          := No_Data;
+      Proxy_Pwd    : String          := No_Data;
+      Timeouts     : Timeouts_Values := No_Timeout;
+      Headers      : Header_List     := Empty_Header_List;
+      Progress     : access procedure
+                       (Total, Sent : Stream_Element_Offset) := null;
+      User_Agent   : String          := Default.User_Agent;
+      HTTP_Version : HTTP_Protocol   := HTTP_Default)
```
TurboGit, line 294: Please remove this option it is just not supported. And I would add a Pre condition on Upload with connection to check that it is HTTP/1 for now.

TurboGit (changes requested):
Minor change !

src/core/aws-client.ads
anisimkov: Done

PR #262, opened 2021-10-08
Rewrite SOAP duration representation

anisimkov:
no-tn-check

Do not need to use monthes and years in duration representation.
Using years and monthes makes the duration representation depend on current
time, which can be different in client and server parts. By the way remove
the error in days calculation.


TurboGit (approved):
Thanks !

PR #263, opened 2021-10-08
Fix connection flow control

anisimkov:
Do not need update connection flow control together with stream
flow control. Connection flow control has to be updated separately.
Fixes file dowmload longer than 2G.

TN: S507-051

### DIFF
--- a/src/core/aws-client-http_utils.adb
+++ b/src/core/aws-client-http_utils.adb
@@ -37,6 +37,7 @@ with Ada.Strings.Maps;
 with AWS.Digest;
 with AWS.Headers.Values;
 with AWS.HTTP2.Connection;
+with AWS.HTTP2.Frame.List;
 with AWS.HTTP2.Frame.Settings;
 with AWS.HTTP2.HPACK.Table;
 with AWS.HTTP2.Message;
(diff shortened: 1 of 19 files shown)


src/http2/aws-http2-frame.ads
```diff
@@ -222,6 +222,8 @@ private
    overriding procedure Adjust (Self : in out Object);
    overriding procedure Finalize (Self : in out Object);
 
+   procedure Dump_Binary (Data : Stream_Element_Array);
```
TurboGit, line 225: to be moved in AWS.Utils as generic.

src/core/aws-client-http_utils.adb
```diff
@@ -417,6 +424,41 @@ package body AWS.Client.HTTP_Utils is
...
+   -- Get_H2_Frame --
+   ------------------
+
+   procedure Get_H2_Frame
+     (Connection : in out HTTP_Connection;
+      Ctx        : in out Server.Context.Object;
+      Stream     : in out HTTP2.Stream.Object)
+   is
+      use AWS.HTTP2;
+      Frame : constant HTTP2.Frame.Object'Class :=
```
TurboGit, line 437: indent :

TurboGit (changes requested):
Just two minor comments, look ok otherwise!

src/core/aws-client-http_utils.adb
anisimkov: Moved

src/http2/aws-http2-frame.ads
anisimkov: Moved

PR #266, opened 2021-10-11
po/http2 upload

TurboGit:


src/core/aws-server-protocol_handler_v2.adb
```diff
@@ -1063,25 +1075,51 @@ begin
       end loop For_Every_Frame;
    end;
 
+   AWS.Status.Set.Free (LA.Stat);
+
 exception
    when Net.Socket_Error =>
-      null;
+      AWS.Status.Set.Free (LA.Stat);
```
anisimkov, line 1082: We don't need to call Free here because we have call to it from Finalize procedure.

PR #273, opened 2021-10-14
Move HTTP2 client context to connection

anisimkov:
TN: S507-051


src/http2/aws-http2-hpack-table.ads
```diff
@@ -71,6 +71,8 @@ package AWS.HTTP2.HPACK.Table is
    procedure Dump (Self : Object);
    --  Dump table content for debug
 
+   procedure Clear (Self : in out Object);
```
TurboGit, line 74: A bit of comment here.

src/http2/aws-http2-frame-settings.adb
```diff
@@ -51,7 +51,7 @@ package body AWS.HTTP2.Frame.Settings is
       Ack      : Boolean := False) return Object
    is
       Len : constant Stream_Element_Count :=
-              Stream_Element_Count (Settings'Length * Payload'Size / 8);
+              Settings'Length * Payload'Size / System.Storage_Unit;
```
TurboGit, line 54: This is not correct to me. 'Size returns bits and Storage_Unit possible depends on the OS, no? The size of the object is imposed by the RFC.

src/core/aws-response.ads
```diff
@@ -449,10 +450,10 @@ private
      & "</BODY></HTML>" & CRLF;
 
    Undefined_Length : constant Content_Length_Type :=
-                        Content_Length_Type (Resources.Undefined_Length);
+                        Resources.Undefined_Length;
 
    type Release_Controller is record
-      Counter      : Natural := 1;
+      Counter      : Utils.Counter (1);
```
TurboGit, line 456: Indeed, lot better !

src/http2/aws-http2-frame-settings.adb
anisimkov: I changed it to Stream_Element'Size. It should be better as we are converting array lengths.

PR #289, opened 2021-11-03
po/fix http2

TurboGit:


anisimkov (approved):
I do not think that 033 should be DEAD on HTTP2, but, we can restore it later.

TurboGit:
> I do not think that 033 should be DEAD on HTTP2, but, we can restore it later.

We'll re discus that but note that there is no Connection header in HTTP/2 so no way to close a connection using this header.

PR #290, opened 2021-11-04
po/fix http2

TurboGit:
Fixes 3 regressions!

### DIFF
--- a/src/core/aws-server-http_utils.adb
+++ b/src/core/aws-server-http_utils.adb
@@ -1361,8 +1361,8 @@ package body AWS.Server.HTTP_Utils is
       C_Stat      : AWS.Status.Data;
       Length      : Response.Content_Length_Type)
    is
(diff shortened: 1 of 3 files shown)


src/core/aws-client.ads
```diff
@@ -70,7 +70,7 @@ package AWS.Client is
    --  Number of time a data is requested from the Server if the first
    --  time fails.
 
-   HTTP_Default : HTTP_Protocol renames HTTPv1;
+   HTTP_Default : HTTP_Protocol renames HTTPv2;
```
anisimkov, line 73: I guess you just forgot to keep HTTPv1 default.

anisimkov (changes requested):

src/core/aws-client.ads
TurboGit: Indeed ! Now fixed.

anisimkov (approved):

PR #291, opened 2021-11-05
Upload control

anisimkov:
Fixes 0188_uplimit in HTTP2 mode

### DIFF
--- a/src/http2/aws-http2-message.ads
+++ b/src/http2/aws-http2-message.ads
@@ -31,6 +31,7 @@ with Ada.Finalization;
 
 with AWS.Headers;
 with AWS.HTTP2.Frame.List;
+with AWS.Messages;
 with AWS.Response;
 with AWS.Server.Context;
 with AWS.Status;
@@ -49,6 +50,8 @@ package AWS.HTTP2.Message is
    use type Response.Data_Mode;
 
    type Object is new Finalization.Controlled with private;
+
+   Undefined : constant Object;
 
    function Is_Defined (Self : Object) return Boolean;
 
@@ -102,8 +105,13 @@ package AWS.HTTP2.Message is
    --  payload depending on the current flow control window. If some more data
    --  is available, More_Frame below will return true.
 
-   function More_Frames (Self : Object) return Boolean;
+   function More_Frames (Self : Object) return Boolean
+     with Pre => Self.Is_Defined;
    --  Returns True if some more data are available and so should be sent
+
+   function Status_Code (Self : Object) return Messages.Status_Code
+     with Pre => Self.Is_Defined;
+   --  Returns message status code
 
 private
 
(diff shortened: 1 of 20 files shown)


src/http2/aws-http2-stream.ads
```diff
@@ -137,6 +141,14 @@ package AWS.HTTP2.Stream is
           Post => Data_Flow'Result = Unknown or else Self.State /= Idle;
    --  Current stream direction
 
+   function Upload_State (Self : Object) return Upload_State_Kind
+     with Pre  => Self.Is_Defined;
```
TurboGit, line 145: 2 spaces before =>

src/http2/aws-http2-stream.adb
```diff
@@ -661,4 +689,14 @@ package body AWS.HTTP2.Stream is
       Self.Flow_Send_Window := Self.Flow_Send_Window + Increment;
    end Update_Flow_Control_Window;
 
+   ---------------------
+   -- Upload_Decision --
+   ---------------------
+
+   procedure Upload_Decision (Self : in out Object; Allow : Boolean) is
+   begin
+      Self.Upload_State := (if Allow then Upload_Accepted
```
TurboGit, line 698: for if expression I usually indent:

```
   (if ..
    then ...
    else ...)
```

src/http2/aws-http2-stream.ads
```diff
@@ -87,7 +90,8 @@ package AWS.HTTP2.Stream is
      (Self  : in out Object;
       Ctx   : in out Server.Context.Object;
       Frame : HTTP2.Frame.Object'Class;
-      Error : out Error_Codes)
+      Error : out Error_Codes;
+      Info  : out Natural)
```
TurboGit, line 94: Looks like Info was some kind of debug data and used only as information into a go-away frame. Can we put this data into the Stream itself and have an function to access this value? This clobber a bit the API for nothing semantic.

Even better would be to have an enumerated type with clear state information. Do you think it can be done?

src/http2/aws-http2-message.ads
```diff
@@ -105,6 +108,8 @@ package AWS.HTTP2.Message is
    function More_Frames (Self : Object) return Boolean;
    --  Returns True if some more data are available and so should be sent
 
+   function Status_Code (Self : Object) return Messages.Status_Code;
```
TurboGit, line 111: Some comments needed and `Pre => Self.Is_Defined` needed. BTW, More_Frames may also need the same pre-condition?

TurboGit (changes requested):
First quick review. Not tested yet.

src/http2/aws-http2-stream.ads
anisimkov: Yes, I did.

anisimkov:
added http2 compatibility to test 0326_big_post by the way.

TurboGit (approved):
Thanks !

PR #294, opened 2021-11-06
Da/http2

anisimkov:


TurboGit:
This does not fix any test on my side. Is that expected?

anisimkov:
> This does not fix any test on my side. Is that expected?

Yes. It fixed ZLib buffer error in the 0213_check_mem_nossl but not fixed the memory leak.

TurboGit (approved):
I see, thanks !

PR #296, opened 2021-11-07
Avoid "attempt to tamper with cursors" exception

anisimkov:
* src/core/aws-server-protocol_handler_v2.adb (Go_Away):
Do not remove stream from S stream set.
It is not necessary because there is exit from For_Every_Frame loop and
return from Protocol_Handler_V2 after Go_Away call.
It was
raised PROGRAM_ERROR :
  AWS.HTTP2.Stream.Set.Maps.Tree_Types.Implementation.TC_Check:
  attempt to tamper with cursors
in AWS error logs on h2spec testing on S.Exclude (Stream_Id).

Part of S507-051.

### DIFF
--- a/src/core/aws-server-protocol_handler_v2.adb
+++ b/src/core/aws-server-protocol_handler_v2.adb
@@ -107,11 +107,6 @@ is
    procedure Queue_Settings_Frame;
    --  Queue server settings frame (default configuration)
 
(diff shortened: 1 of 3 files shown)


TurboGit (approved):
Nice! Works for me, thanks!

PR #303, opened 2021-11-10
Da/http2

anisimkov:


TurboGit (changes requested):
Nice but it introduces a regression on 0043_check_mem.

TurboGit (approved):
Looks good, and no regression on my side now. Thanks !

PR #307, opened 2021-11-17
Fix regression after HTTP2 activated by default

anisimkov:
Regression in 0279_broken_status_line.

* src/core/aws-server-protocol_handler.adb:
Use GoAway frame response only in HTTP/2 protocol.

Part of S507-051.


TurboGit (approved):
Thanks !

PR #310, opened 2021-11-22
Force clean without close the socket when possible

anisimkov:
We can release task and do not close the socket on hard loaded server at
time when task waiting for next client request.
Minor optimisation.

### DIFF
--- a/src/core/aws-server.ads
+++ b/src/core/aws-server.ads
@@ -35,7 +35,9 @@ with AWS.Config;
 with AWS.Default;
 with AWS.Dispatchers;
 with AWS.Exceptions;
+with AWS.Net.Poll_Events;
 with AWS.Net.SSL;
+with AWS.Net.Std;
 with AWS.Response;
 with AWS.Status;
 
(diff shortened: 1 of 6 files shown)


src/core/aws-server.ads
```diff
@@ -334,6 +336,9 @@ private
 
    type Slot is record
       Sock                  : Socket_Access;
+      Wait_For_Data         : aliased Net.Poll_Events.Set (2);
+      Wait_Breaker          : Net.Std.Socket_Type;
```
TurboGit, line 340: Breaker?

anisimkov: Yes, to break the waiting call.

TurboGit: A signaling socket? At least please put a little comment here as it was not trivial to understand when I have reviewed the changes.

anisimkov: Added comment at the end of record declaration. Removed obsolete comments there by the way.

src/core/aws-client.ads
```diff
@@ -70,7 +70,7 @@ package AWS.Client is
    --  Number of time a data is requested from the Server if the first
    --  time fails.
 
-   HTTP_Default : HTTP_Protocol renames HTTPv1;
+   HTTP_Default : HTTP_Protocol renames HTTPv2;
```
TurboGit, line 73: Unwanted change.

TurboGit (changes requested):

src/core/aws-client.ads
anisimkov: Opps, .... fixed.

TurboGit (approved):
Looks good now. Thanks.

PR #311, opened 2021-11-27
Minor optimization

anisimkov:
Send messages without body even if zero flow control window.
Detect earlier frame arrival and reading it before send everything from
Answers container.

Part of S507-051.


TurboGit (approved):
Looks good, thanks!

Issue #316, opened 2021-12-26
CONSTRAINT_ERROR for byte ranges


reznikmm:
Hello,

Trying to fetch a video from AWS I get CONSTRAINT_ERROR in `Send_Range` procedure of aws-server-http_utils.adb:
```
            Send
              (Messages.Content_Range_Token & ": bytes "
               & Utils.Image (Natural (R_First)) & "-"
               & Utils.Image (Natural (R_Last))   ---  <<<
               & "/" & Utils.Image (Natural (Length))   ---  <<<
               & CRLF
               & Messages.Content_Length (R_Length) & CRLF & CRLF);
 ```

because the length of the file is more then Natural'Last (which is just 2Gb on my OS).

Could you fix that?

TurboGit:
That is fixed now.

PR #318, opened 2021-12-28
Fix possible Constraint_Error on integer overflow

anisimkov:
Do not convert Stream_Element_Offset to Integer on call to AWS.Utils.Image,
use appropriate AWS.Utils.Image instead. Fixes possible similar issue as
described in #316 PR.


TurboGit (approved):
Thanks!

PR #325, opened 2022-02-05
Improve SSL_DYNAMIC option explanation

anisimkov:
TN: UB30-003

### DIFF
--- a/shared.gpr
+++ b/shared.gpr
@@ -139,10 +139,9 @@ abstract project Shared is
    --  Common options used for the Debug and Release modes
 
    Debug_Options :=
-     ("-g", "-gnata", "-gnatVa", "-gnatQ", "-gnato", "-gnatwe", "-Wall");
+     ("-g", "-gnata", "-gnatVa", "-gnatQ", "-gnato", "-Wall", "-gnatwe");
 
-   Release_Options :=
-     ("-O2", "-gnatn");
+   Release_Options := ("-O2", "-gnatn");
 
    package Compiler is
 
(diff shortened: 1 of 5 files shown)


shared.gpr
```diff
@@ -154,6 +153,12 @@ abstract project Shared is
             for Default_Switches ("C")
               use ("-g", "-Wno-implicit-function-declaration");
 
+            for Switches ("ssl-thin.ads") use Common_Options & Debug_Warnings;
+            --  The file ssl-thin.ads is generated from ssl-thin__openssl.ads
+            --  or from ssl-thin__gnutls.ads and reformatted with gnatpp, but
+            --  gnatpp unable to split long line of aspect list, see V117-043.
+            --  It is because we do not use "-gnatwe" option to compile it.
```
TurboGit, line 160: It is why we..

shared.gpr
```diff
@@ -138,11 +138,10 @@ abstract project Shared is
      & Global_Options & Instrument_Options;
    --  Common options used for the Debug and Release modes
 
-   Debug_Options :=
-     ("-g", "-gnata", "-gnatVa", "-gnatQ", "-gnato", "-gnatwe", "-Wall");
+   Debug_Warnings := ("-g", "-gnata", "-gnatVa", "-gnatQ", "-gnato", "-Wall");
+   Debug_Options  := Debug_Warnings & "-gnatwe";
```
TurboGit, line 142: Instead of splitting in two options I think you could have inhibited the -gnatwe option for ssl-thin.ads below.

Using -gnat-N or -gnat-m (only line length disabled).

TurboGit (commented):
Have you tried the only disabling some checks for the generated ssl-thin.ads?

shared.gpr
anisimkov: I increased the max line length for ssl-thin.ads.

TurboGit (approved):
Looks good, thanks!

PR #335, opened 2022-05-23
add support for so_closexec option

TurboGit:


TurboGit:
I have no idea for testing this in the regression test for now.

anisimkov (approved):

PR #338, opened 2022-06-17
Fix internal WebSocket message buffer to avoid a constraint error.

TurboGit:
Fixes V617-023.

### DIFF
--- a/src/core/aws-net-websocket-registry.adb
+++ b/src/core/aws-net-websocket-registry.adb
@@ -1,7 +1,7 @@
 ------------------------------------------------------------------------------
 --                              Ada Web Server                              --
 --                                                                          --
---                     Copyright (C) 2012-2021, AdaCore                     --
+--                     Copyright (C) 2012-2022, AdaCore                     --
 --                                                                          --
 --  This library is free software;  you can redistribute it and/or modify   --
 --  it under terms of the  GNU General Public License  as published by the  --
@@ -1401,9 +1401,7 @@ package body AWS.Net.WebSocket.Registry is
       Timeout      : Duration := Forever;
       Asynchronous : Boolean := False)
    is
-      A : String (Natural (Message'First) .. Natural (Message'Last))
-         with Import;
-      for A'Address use Message'Address;
+      A : String (1 .. Message'Length) with Import, Address => Message'Address;
       pragma Compile_Time_Error
          (Character'Size /= Stream_Element'Size,
           "A character size is not the size of a byte");


src/core/aws-net-websocket-registry.adb
```diff
@@ -1401,7 +1401,7 @@ package body AWS.Net.WebSocket.Registry is
       Timeout      : Duration := Forever;
       Asynchronous : Boolean := False)
    is
-      A : String (Natural (Message'First) .. Natural (Message'Last))
+      A : String (1 .. Natural (Message'Length))
```
anisimkov, line 1404: No need conversion to Natural here. 'Length compatible with all types as I can remember.

anisimkov (changes requested):

src/core/aws-net-websocket-registry.adb
briot: In fact, this piece of code is potentially dangerous. What if you receive a message more than 2MB ? you'll blow up the stack and end up with Stack_Overflow.

TurboGit: > No need conversion to Natural here. 'Length compatible with all types as I can remember.

Indeed. Now fixed.

TurboGit: > In fact, this piece of code is potentially dangerous. What if you receive a message more than 2MB ? you'll blow up the stack and end up with Stack_Overflow.

I don't think it is the case here as the variable A is imported and overlaid on top of current parameter. So at the end it does not take any additional byte on the stack.

briot: I do not understand the "is imported" notion. You are declaring a new array on the stack. It has nothing to do with the parameters that the subprogram receives (except for looking at the bounds). So it will blow up. I actually fixed a similar case in my local copy of AWS (I don't remember whether it was this specific piece of code or another place, though). Does the AWS testsuite have a test where it sends a 5MB message, and another where it receives a 5MB message ?

TurboGit: It is not only imported but it has an Address clause.

TurboGit: @briot : You can test it by yourself, even 150 Mb is passing without an exception. There is nothing put on the stack because of the Address clauses.

```
with Ada.Text_IO;
with Ada.Streams;

procedure Tmem is

   use Ada.Text_IO;
   use Ada.Streams;

   procedure Send (P : String) is
   begin
      for C of P loop
         if C /= 'x' then
            Put_Line ("error!");
         end if;
      end loop;
   end Send;

   procedure Send (P : Stream_Element_Array) is
      A : String (1 .. P'Length) with Address => P'Address;
   begin
      Send (A);
   end Send;

   type SA is access Stream_Element_Array;

   P : SA := new Stream_Element_Array (1 .. 150 * 1_024 * 1_024);

begin
   for I in P'range loop
      P (I) := Character'Pos ('x');
   end loop;

   Send (P.all);

   Put_Line ("All good !");
end Tmem;
```

anisimkov (approved):

PR #340, opened 2022-06-22
Adapt to libadalang API changes

thvnx:
Libadalang got rid of `IndexConstraint` and `DiscriminantConstraint`
nodes, replacing them by a single `CompositeConstraint`, with new
`p_is_index_constraint` and `p_is_discriminant_constraint` properties.

TN: U526-014
TN: V622-016


TurboGit:
Hello, I am supposed to merge this into master now?

thvnx:
I don't think so, you'll just need it for the next release. I guess that until you don't change the version of libadalang the stable version of aws is using, you shouldn't merge this into master.

yakobowski:
@TurboGit no you don't. (And should not, as this would break some of our local builds.) This will be merged into master at the next global AdaCore "stale bump".

PR #341, opened 2022-06-28
Increase length of certificate key

anisimkov:
OpenSSL 1.1.1f do not want to worl with too short certificate key.


TurboGit (changes requested):
Ok, but can you make the certificate with a longer validity. It was until 2032 and the new version will expire on 2022. Well actually it is already invalid (Feb 2022). TIA.

anisimkov:
I did not cerate it. I just

anisimkov:
Expiration date increased

PR #352, opened 2022-11-24
soap std vector & generate based on templates

TurboGit:
This is a large PR that has started as adding support for standard Ada.Containers.Vector for arrays in SOAP support.

Seeing how difficult it was to get this right the generation of Ada code for the stubs, skeletons, callbacks & main has been migrated to templates file (based on Template_Parser engine). This makes the code easier to read but there is still some room for simplification.

This is a draft as some more clean-up must be done.


leogermond (changes requested):
This requires a TN

TurboGit:
@leogermond : When you say this requires a TN, you speak about which commit?

leogermond:
> @leogermond : When you say this requires a TN, you speak about which commit?

I'm speaking about the PR as a whole, unless some commit in particular can be tracked back to another ticket.

leogermond:
TN is Re: [V531-036]

leogermond (dismissed):

leogermond (approved):

PR #368, opened 2023-08-04
Make 'make setup' macOS-compatible.

simonjwright:
'makefile.conf' uses the sed switch '-s', which is not available on macOS; macOS sed is derived from BSD sed, and (according to https://man.freebsd.org/cgi/man.cgi?sed(1)) FreeBSD sed 13.0 doesn't support -s either.

Since the GNU sed -s (--separate) only applies if there are multiple input files, and in each case here there's only one input file, the -s can be removed without penalty.

It would be great if this change could be fed though to the alire version (that’d be v23.0.1, I suppose).

### DIFF
--- a/makefile.conf
+++ b/makefile.conf
@@ -147,13 +147,13 @@ PROCESSORS	= 0
 # HTTP2 server activated
 #
 SERVER_HTTP2 = $(strip \
-                 $(shell sed -ns 's/HTTP2_Activated.*:= *\([a-zA-Z]*\);/\1/p' \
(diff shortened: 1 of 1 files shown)


simonjwright:
I should add, this doesn’t stop the build: it’s just worrying.
```
sed: illegal option -- s
usage: sed script [-Ealnru] [-i extension] [file ...]
        sed [-Ealnu] [-i extension] [-e script] ... [-f script_file] ... [file ...]
```

TurboGit:
This is now merged, thanks!

PR #392, opened 2025-05-31
fix: memory leak in SMTP server Get_Message function

younes-io:
Found a small memory leak in the SMTP server: `Get_Message` creates a socket copy but forgets to free it. The actual socket is fine (it's ref-counted), just the wrapper that leaks.

Added the missing `Net.Free` call + exception handler to clean up properly. Nothing fancy, just following the same cleanup pattern used elsewhere in AWS.

### DIFF
--- a/src/extended/aws-smtp-server.adb
+++ b/src/extended/aws-smtp-server.adb
@@ -220,7 +220,13 @@ package body AWS.SMTP.Server is
       Messages.Set.Set_Body (Message, To_String (Message_Body));
       Messages.Set.Headers (Message, Headers);
 
+      Net.Free (S);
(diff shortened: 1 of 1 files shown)


ogorodnik (approved):

ogorodnik:
@younes-io Hello, we moved your patch to our repository, which mirrors the code here.  I'll close this PR if GitHub does not close it automatically (in general, it does when it sees the MR copied).

ogorodnik:
@younes-io You can create MRs here, we will take care of them.

ogorodnik:
@younes-io You can create MRs here, we will take care of them.

